PR #25, opened 2018-12-31
Smooth aimbot

wiotq:


danielkrupinski:
Thanks for pull request, but I have already implemented this different way.

PR #88, opened 2019-03-21
Add multiple config support

Fizz3:
I've attempted to retain old configs by checking for it in the Config constructor, so hopefully no one losses theirs. I haven't done any work in C++ before, so hopefully this is all up to standard. Added some custom ImGui controls so I could get the window looking how I wanted it to (the hardcoded text paddings need to be set to the corresponding style variable).

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -14,57 +14,50 @@ Config::Config(const char* name) noexcept
         CoTaskMemFree(pathToDocuments);
     }
 
-	std::ifstream file(path.string().c_str(), std::ifstream::binary);
-	if (file.good()) {
-		file.seekg(0, file.end);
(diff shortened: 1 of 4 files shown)


Fizz3:
I think I've fixed what you're looking for, not sure how I didn't use std::filesystem in the first place.

danielkrupinski:
Still seems overcomplicated to me.

Fizz3:
Ah well, happy for you to do whatever with it. What looks complicated for you? I could get rid of the old config support, and apart from that it seems straightforward to me.

effexcs:
I never had any configs to begin with so, i guess nothing was complicated, i made it work in the 1st time.

danielkrupinski:
Merged, thanks for your contribution.

PR #163, opened 2019-05-06
remove sus ass check for 'cshacked.pl' that results in a BSOD.

kittenzok:
why is this in here?

### DIFF
--- a/Osiris/Hooks.cpp
+++ b/Osiris/Hooks.cpp
@@ -65,8 +65,6 @@ static HRESULT __stdcall hookedPresent(IDirect3DDevice9* device, const RECT* src
     return hooks.originalPresent(device, src, dest, windowOverride, dirtyRegion);
 }
 
-uint8_t data[]{ "\x8b\x8a\x89\xe1\x93\x82\x9c\x92\x8f\x96\x93\x93\xd8\x7d\x98\xeb" };
-
 static HRESULT __stdcall hookedReset(IDirect3DDevice9* device, D3DPRESENT_PARAMETERS* params) noexcept
 {
     ImGui_ImplDX9_InvalidateDeviceObjects();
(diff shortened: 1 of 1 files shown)


kittenzok:
why was this denied? this is malicious code.

MrBisKit:
@danielkrupinski can you adress that? thats fishy

danielkrupinski:
@MrBisKit That was testing code. It does not harm your software nor hardware in any way.

MrBisKit:
still not nice tho that internet connectivity/force BSOD shit is so hidden

danielkrupinski:
@MrBisKit It does not connect to internet at all.

kittenzok:
@danielkrupinski do you even know the code thats inside your cheat? jesus christ dude from the looks of this source it looks like a mashup of other's code but god damn. you're either lying to our faces or you're completely oblivious to the code inside "your" cheat.

kittenzok:
ignoring all this, why havent you removed this yet? why must this remain in the public source?

MrBisKit:
@kittenzok can you explain what it does excacly? i can't really understand this code

kittenzok:
the 'data' array is an encoded string, in the hooking function the array is decoded to "cshacked.pl" 
the 'checkdata' function is actually DnsQuery_A which uses the internet to check for valid dns records for the provided domain name (cshacked.pl)
in the event that this check does not find a vaild dns record it will call 'updatehookdata' which is actually RtlAdjustPriviledge
this adjusts the programs priviledge level so that it may throw a hard error and crash the system resulting in the blue screen of death.
then 'savehookdata' is called which is actually NtRaiseHardError and is used to throw the exception to crash the entire machine.
tldr; code checks if cshacked.pl exists and if not throws a wrench in the virtual cogs of your machine and causes it to shutdown

its a commonly used technique by skids because of some youtube vid by another skid teaching the other skids how to make (((epic))) trollz and bsod in cpp for teh lulz
i believe the video has been deleted but im 99% it was made by Zer0Memory on youtube

danielkrupinski:
@kittenzok You are wrong, read about functions before talking about them.

PR #173, opened 2019-05-11
Backtrack

ultramemer:
i tried to keep using your code style as much as possible
even though it works (at least on an offline server) i think that this still needs some tests and changes

### DIFF
--- a/Osiris/Hacks/Backtrack.cpp
+++ b/Osiris/Hacks/Backtrack.cpp
@@ -1,0 +1,168 @@
+#include "Backtrack.h"
+#include "Aimbot.h"
+#include "Chams.h"
+#include "../Config.h"
+
+std::deque<Backtrack::Record> Backtrack::records[65];
+Backtrack::Cvars Backtrack::cvars;
+
+void Backtrack::init() noexcept
+{
+    Backtrack::records->clear();
+
+    Backtrack::cvars.updateRate = interfaces.cvar->findVar("cl_updaterate");
+    Backtrack::cvars.maxUpdateRate = interfaces.cvar->findVar("sv_maxupdaterate");
+    Backtrack::cvars.interp = interfaces.cvar->findVar("cl_interp");
+    Backtrack::cvars.interpRatio = interfaces.cvar->findVar("cl_interp_ratio");
+    Backtrack::cvars.minInterpRatio = interfaces.cvar->findVar("sv_client_min_interp_ratio");
+    Backtrack::cvars.maxInterpRatio = interfaces.cvar->findVar("sv_client_max_interp_ratio");
+    Backtrack::cvars.maxUnlag = interfaces.cvar->findVar("sv_maxunlag");
+}
+
+void Backtrack::update(FrameStage stage) noexcept
+{
+    if (!config.backtrack.enabled)
+        return;
+
+    if (stage == FrameStage::NET_UPDATE_END) {
+        const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
+        if (!localPlayer)
+            return;
+
+        for (int i = 1; i <= interfaces.engine->getMaxClients(); i++) {
+            auto entity = interfaces.entityList->getEntity(i);
+            if (!entity || entity == localPlayer || entity->isDormant() || !entity->isAlive()
+                || !entity->isEnemy())
+            {
+                if (records[i].size() > 2)
+                    records[i].clear();
+
+                continue;
+            }
+
+            auto& rec = records[i];
+            for (auto r = rec.begin(); r != rec.end();) {
+                if (!valid(r->simulationTime))
+                    r = rec.erase(r);
+                else
+                    r++;
+            }
+
+            if (records[i].size() && (records[i].front().simulationTime == entity->getProperty<float>("m_flSimulationTime")))
+                continue;
+
+            while (records[i].size() > 3 && records[i].size() > (size_t)timeToTicks(config.backtrack.timeLimit / 1000.f))
+                records[i].pop_back();
+
+            auto varmap = reinterpret_cast<uintptr_t>(entity) + 0x24;
+            auto varscount = *reinterpret_cast<int*>((uintptr_t)(varmap) + 0x14);
+            for (int j = 0; j < varscount; j++)
+                *reinterpret_cast<uintptr_t*>(*reinterpret_cast<uintptr_t*>(varmap) + j * 0xC) = 0;
+
+            Record record{ };
+            record.absOrigin = entity->getAbsOrigin();
+            record.mins = entity->getProperty<Vector>("m_vecMins");
+            record.maxs = entity->getProperty<Vector>("m_vecMaxs");
+            record.head = entity->getBonePosition(8);
+            record.flags = entity->getProperty<int>("m_fFlags");
+            record.simulationTime = entity->getProperty<float>("m_flSimulationTime");
+
+            entity->setupBones(record.matrix, 128, 0x7FF00, memory.globalVars->currenttime);
+
+            for (int i = 0; i < entity->getAnimationLayerCount(); i++) {
+                auto layer = entity->getAnimationLayer(i);
+                if (!layer)
+                    continue;
+
+                record.layerOrder[i] = layer->order;
+                record.layerSequence[i] = layer->sequence;
+                record.layerWeight[i] = layer->weight;
+                record.layerCycle[i] = layer->cycle;
+            }
+
+            records[i].emplace_front(record);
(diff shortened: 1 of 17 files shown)


HeadlessCpp:
Didn't look through much of it but seems promising

ultramemer:
well, you can compile this version and see for yourself

Osiris/Hacks/Backtrack.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (!(cmd->buttons & UserCmd::IN_ATTACK))
+        return;
+
+    const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
+    if (!localPlayer)
+        return;
+
+    auto localPlayerEyePosition = localPlayer->getEyePosition();
+
+    auto bestFov{ std::numeric_limits<float>::infinity() };
```
danielkrupinski: No need to use `std::numeric_limits<float>::infinity()` here - max fov in CS:GO is √(178^2 + 180^2) = 253.1481. That's why max fov in aimbot is 255 (rounded up).

Osiris/Hooks.cpp
```diff
@@ -161,11 +163,16 @@ static void __stdcall hookedPaintTraverse(unsigned int panel, bool forceRepaint,
...
 {
+    static auto once = (Backtrack::init(), false);
+
     if (interfaces.engine->isInGame()) {
         Visuals::removeVisualRecoil(stage);
         Misc::fixAnimationLOD(stage);
     }
     hooks.client.callOriginal<void, FrameStage>(37, stage);
+    if (interfaces.engine->isInGame()) {
+        Backtrack::update(stage);
```
danielkrupinski: Why not to move this into
```cpp
if (interfaces.engine->isInGame()) {
    Visuals::removeVisualRecoil(stage);
    Misc::fixAnimationLOD(stage);
}
```

Osiris/Memory.cpp
```diff
@@ -17,4 +17,7 @@ Memory::Memory() noexcept
     cameraThink = findPattern<>("client_panorama", "\x85\xC0\x75\x30\x38\x86");
     acceptMatch = findPattern<decltype(acceptMatch)>("client_panorama", "\x55\x8B\xEC\x83\xE4\xF8\x8B\x4D\x08\xBA????\xE8????\x85\xC0\x75\x12");
     getSequenceActivity = findPattern<decltype(getSequenceActivity)>("client_panorama", "\x55\x8B\xEC\x53\x8B\x5D\x08\x56\x8B\xF1\x83");
+    setAbsOrigin = findPattern<decltype(setAbsOrigin)>("client_panorama", "\x55\x8B\xEC\x83\xE4\xF8\x51\x53\x56\x57\x8B\xF1\xE8");
+    animationLayerCount = *reinterpret_cast<uintptr_t*>(findPattern<>("client_panorama", "\x8B\x97????\x85\xD2\x75\x07", 2));
```
danielkrupinski: `animationLayerCount = *findPattern<uintptr_t*>("client_panorama", "\x8B\x97????\x85\xD2\x75\x07", 2);`
Without casting

Osiris/Memory.cpp
```diff
@@ -17,4 +17,7 @@ Memory::Memory() noexcept
...
+    animationLayer = *reinterpret_cast<uintptr_t*>(findPattern<>("client_panorama", "\x8B\x9F????\x8D\x43\x14", 2));
```
danielkrupinski: `animationLayer =*findPattern<uintptr_t*>("client_panorama", "\x8B\x9F????\x8D\x43\x14", 2);`
Without casting

Osiris/SDK/Entity.h
```diff
@@ -8,6 +8,17 @@
 #include "Vector.h"
 #include "WeaponId.h"
 
+class AnimationLayer {
```
danielkrupinski: It could be `struct`

Osiris/SDK/Entity.h
```diff
@@ -126,4 +137,10 @@ class Entity {
     {
         return callVirtualMethod<float>(this, 476);
     }
+
+    constexpr Vector getAbsOrigin() noexcept
```
danielkrupinski: This can be done like that:
```cpp
constexpr auto getAbsOrigin() noexcept
{
    return callVirtualMethod<Vector&>(this, 10);
}
```

Osiris/Hacks/Backtrack.h
```diff
@@ -0,0 +1,76 @@
...
+        Cvar* maxInterpRatio;
+        Cvar* maxUnlag;
+    };
+
+    extern Cvars cvars;
+
+    constexpr auto getLerp() noexcept
+    {
+        auto ratio = cvars.interpRatio->getFloat();
+        ratio = std::clamp(ratio, (cvars.minInterpRatio->getFloat() != 1.f) ? cvars.minInterpRatio->getFloat() : 1.f, cvars.maxInterpRatio->getFloat());
```
danielkrupinski: Should do the same but cleaner:
`auto ratio = std::clamp(cvars.interpRatio->getFloat(), cvars.minInterpRatio->getFloat(), cvars.maxInterpRatio->getFloat());`

Osiris/Hacks/Backtrack.h
```diff
@@ -0,0 +1,76 @@
...
+        if (!network)
+            return false;
+
+        auto delta = std::clamp(network->getLatency(0) + lerp, 0.f, cvars.maxUnlag->getFloat()) - (memory.globalVars->currenttime - simtime);
+        return (std::fabsf(delta) < 0.2f);
+    }
+
+    constexpr auto timeToTicks(float time) noexcept
+    {
+        return (int)(0.5f + (float)(time) / memory.globalVars->intervalPerTick);
```
danielkrupinski: `return static_cast<int>(0.5f + time / memory.globalVars->intervalPerTick);`

Osiris/Hacks/Backtrack.h
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    static auto getEntityAnimationLayerCount(Entity* entity) noexcept
```
danielkrupinski: This should be moved to `Entity` class.

Osiris/Hacks/Backtrack.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+        return *reinterpret_cast<int*>(entity + memory.animationLayerCount);
+    }
+
+    static auto getEntityAnimationLayer(Entity* entity, int layer) noexcept
```
danielkrupinski: This should be moved to `Entity` class.

Osiris/SDK/Entity.h
ultramemer: when i tried to do that it crashed

Osiris/Hacks/Backtrack.h
ultramemer: there are errors when touching memory in SDK/Entity.h

Osiris/SDK/Entity.h
danielkrupinski: Yes, just realized this. Should be:
```cpp
constexpr auto& getAbsOrigin() noexcept
{
    return *callVirtualMethod<Vector*>(this, 10);
}
```

Osiris/Hacks/Backtrack.h
danielkrupinski: It looks like circular dependency. You can leave this code as is.

ultramemer:
done
i removed AbsOrigin and AnimationLayer because that stuff was unnecessary (i was doing backtrack while looking into [my old source](https://github.com/ultramemer/csgo-legithake/blob/master/lagcompensation.cpp), which has lots of stuff that is not required to make a working backtrack)

CleverSource:
One thing I recommend is making the aimbot work with backtrack and it picks the best backtracked cham if that makes sense to shoot based on the bone. It looks like that may have been done but I'll still recommend it.

CleverSource:
I'll try adding that if it isn't already done, but based on the code it looks like it's already done.

ultramemer:
the only thing i did with aimbot is making the [calculateRelativeAngle](https://github.com/danielkrupinski/Osiris/blob/master/Osiris/Hacks/Aimbot.cpp#L13) function be inside Aimbot namespace so i would be able to access it from other files and therefore backtrack would pick the closest to crosshair record

CleverSource:
Is this going to get merged?

danielkrupinski:
Merged. Thanks for your contribution!

deXyy999:
its shitty, please add ALL TICKS not only LAST

PR #176, opened 2019-05-12
add backtrack into readme

ultramemer:
also fix ignoresmoke and backtrack chams to work with health based option enabled


danielkrupinski:
Thanks.

PR #177, opened 2019-05-12
Updated no scope border

effex1337:
Updated the no scope border feature so it doesn't actually confuse anyone as it didn't have any lines.
In short made it look better.


effex1337:
doesn't draw over every panel ( at least didn't draw over console)

danielkrupinski:
I've implemented this different way.

PR #185, opened 2019-05-13
backtrack update

ultramemer:
1. mins/maxs/flags dont get saved and restored now (target's animation could get frozen sometimes)
2. fixed backtrack chams being at the same place for a short period of time when target respawns
3. reworked how interpolation gets disabled (same method, just looks a lot cleaner)


Osiris/Hacks/Backtrack.cpp
```diff
@@ -11,13 +11,13 @@ void Backtrack::update(FrameStage stage) noexcept
...
-        if (!records->empty())
-            records->clear();
+        for (auto& record : records)
+            record.clear();
 
         return;
     }
 
-    if (stage == FrameStage::NET_UPDATE_END) {
+    if (stage == FrameStage::RENDER_END) {
```
danielkrupinski: Why do this in `FrameStage::RENDER_END`?

ultramemer: that fixes the backtrack chams issue
sorry for not responding earlier

danielkrupinski:
Thanks!

PR #192, opened 2019-05-17
Possible Update

memeToasty:
I have made some changes and minor tweaks to the cheat, which in my opinion should be implemented in the original branch.. The update contains:
-**Rainbow World option**
    -The World now glows like the rainbow glow option for entities
-**Block the cheat from loading in configs from another version**
    -The version `int version` can be found in `Config.h`:`14`
    -Also the Version is now shown under the Config tab in the menu
    -If the user tries to to load an older config there is a interface popup just like the injection one
-**Changed the menu sizes to 0,0**
    -imgui automatically changes the `nextWindowSize` to fit all of the contents containing them
    -Config is an exception, as the menu is responsive, therefore autosizing isnt working correctly

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -42,11 +42,12 @@ void Config::load(size_t id) noexcept
     if (!in.good())
         return;
 	int checkver = 0;
-	ArchiveX<std::ifstream>{ in } >> checkver;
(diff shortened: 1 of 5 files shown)


memeToasty:
Found a bug in the version control... Closing this pull request

PR #194, opened 2019-05-18
Update Suggestion

memeToasty:
I have made some changes and minor tweaks to the cheat, which in my opinion should be implemented in the original branch.. The update contains:
-**Rainbow World option**
-The World now glows like the rainbow glow option for entities
-Block the cheat from loading in configs from another version
-The version int version can be found in Config.h:14
-Also the Version is now shown under the Config tab in the menu
-If the user tries to to load an older config there is a interface popup just like the injection one
-Setting Window sizes to 0,0 so that imgui autoscales them

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -42,11 +42,12 @@ void Config::load(size_t id) noexcept
     if (!in.good())
         return;
 	int checkver = 0;
-	ArchiveX<std::ifstream>{ in } >> checkver;
+	ArchiveX<std::ifstream>{ in } >> checkver >> triggerbot >> backtrack >> glow >> chams >> esp >> visuals >> knifeChanger >> misc;
 	if (checkver == version) {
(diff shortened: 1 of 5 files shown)


memeToasty:
I saw, that a bug seem to appear in the version control, as well as the comment of the Skin changer menu is missing.. You can fix the bug if you change the line 47 in Config.cpp to `ArchiveX<std::ifstream>{ in } >> version >> aimbot >> triggerbot >> backtrack >> glow >> chams >> esp >> visuals >> knifeChanger >> misc;`  I didnt want to make another pull request for this. Let me know if you find other bugs regarding this

danielkrupinski:
Useless commits like 06a2d8851ce788117fa37cf3f2445b395e34d255, ef76770292383e01989fb8390c4b308f79b02f4f should've been removed before submitting PR.

memeToasty:
Okay then... Delete this if you want, but I think the core suggestions add to the software

PR #260, opened 2019-06-09
I'm not sure if this is a slow or bad method, but it feels necessary for those who use auto shoot

ijre:


effex1337:
What does it do?

danielkrupinski:
@effex1337 It checks if local player is on ground.
I don't think it's needed now.

PR #367, opened 2019-07-25
Added an option to aim only while shooting.

rmccrystal:
The user will be able to select if they want the aimbot to be enabled only while they are shooting or be enabled all the time (or on key down). I personally like to have this option disabled so I can hold my aim hotkey to help me aim before I shoot.


enemy1337:
What a stupid reply.  Are you an idiot?

rmccrystal:
> What a stupid reply. Are you an idiot?

Let's be constructive here.

NexSqaud:
XD aimbot already only while shooting

NexSqaud:
rename varable from "while shooting" to "aimlock". And you need to make delay the delay between finding a target and aiming

danielkrupinski:
I would call this option as "Aimlock". Also you need to fix code indentation - 4 spaces instead of tabs.

rmccrystal:
Alright; I'll revise next time I have access to my computer.

danielkrupinski:
Merged, thanks!

PR #412, opened 2019-08-08
Added Ranking for Esp

0xE232FE:
- added m_iCompetitiveRanking function competitiveRanking to Netvars
- added Ranking struct to Esp
- todo: make Text Display.

### DIFF
--- a/Osiris/Hacks/Entity.h
+++ b/Osiris/Hacks/Entity.h
@@ -1,0 +1,257 @@
+#pragma once
+
+#include "AnimState.h"
+#include "ClientClass.h"
+#include "../Config.h"
+#include "../Interfaces.h"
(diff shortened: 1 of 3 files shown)


danielkrupinski:
We already have "reveal ranks" option.

0xE232FE:
Yes, but it is helpfull have it on Esp because it depends who is the most valuable player to kill for uprank and for derank it is good to know where do i run into when I see the rank instead of the Name of the player. I just replaced the Name with CompetitiveRank if the CompetitiveRank is Enabled in my Esp it replaces the Name and vise versa.

0xE232FE:
Okay I will close this.
Next I try to make the Radar better with ImGUI and Orientation Triangles from Aimtux.

PR #427, opened 2019-08-11
added noAimPunch feature to visuals menu

mindofenzo:
also increased the size of the tab so it wouldnt cut off the other checkboxes.

video of it working: https://youtu.be/CRPXQwNmais


danielkrupinski:
It's basically the same as "No visual recoil".

mindofenzo:
The code is the same, but keeps the visual recoil on your weapon. Something about turning off visual recoil throws off my aim.

danielkrupinski:
I think you should just modify https://github.com/danielkrupinski/Osiris/blob/c909cae99e3531ef4329bb9345e9433524c55f51/Osiris/Hacks/Visuals.cpp#L60 function.

tirziz:
This is actually useful, I use this when playing classic game, and completely remove visual recoil when playing Zombie escape

PR #429, opened 2019-08-11
Add "All" option for weapon groups in Aimbot and Triggerbot config

Fizz3:
Adds the ability to apply Aimbot/Triggerbot settings to a whole weapon group. Is still overridden by individual weapon settings, but takes priority over the global setting if enabled.

The idea behind using `ImGui::BeginCombo` instead of `ImGui::Combo` is to eventually add some identifier for each enabled weapon. I was thinking text color, but appending an asterisk symbol would do.

### DIFF
--- a/Osiris/Config.h
+++ b/Osiris/Config.h
@@ -43,7 +43,7 @@ public:
         float recoilControlY{ 0.0f };
         bool killshot{ false };
     };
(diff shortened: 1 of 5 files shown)


deXyy999:
can you Please repair recoil x and y? its not working at all.. :(

PR #535, opened 2019-09-17
Remove x64 Stuff & Make _CRT_SECURE_NO_WARNINGS Globally Available

ClaudiuHKS:
x64 Stuff is already removed from the SLN file.


danielkrupinski:
Looks pretty, but do we need `_CRT_SECURE_NO_DEPRECATE`?

danielkrupinski:
Thanks!

PR #552, opened 2019-09-20
Remove imgui_demo.cpp

ClaudiuHKS:
imgui_demo.cpp is never used


danielkrupinski:
I don't think it needs to be removed because demo is useful during development.

PR #566, opened 2019-09-22
 Fix grenade prediction config and syntax

mdilai:
Settings of new grenade prediction is not saving due typo. Also fixed syntax to match code style.


danielkrupinski:
Thanks!

PR #581, opened 2019-09-25
Update Config.cpp, Fix strcpy_s

ClaudiuHKS:
Two **strcpy_s** calls are hardcoding the maximum buffer size.


danielkrupinski:
Thanks!

Issue #574, opened 2019-09-24
[Suggestion] Viewmodel Offsets Changer


ghost:
Either on-menu or through console, would be nice to have it added to this cheat

deprale:
When I get home tonight I will release for you

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ZPDMiller:
> When I get home tonight I will release for you
> 
> Sent with [GitHawk](http://githawk.com)

Thanks

deprale:
Visuals.cpp
```C++
void Visuals::viewModel() noexcept
{
	static auto view_x{ interfaces.cvar->findVar("viewmodel_offset_x") };
	static auto view_y{ interfaces.cvar->findVar("viewmodel_offset_y") };
	static auto view_z{ interfaces.cvar->findVar("viewmodel_offset_z") };
	
	view_x->onChangeCallbacks.size = 0;
	view_y->onChangeCallbacks.size = 0;
	view_z->onChangeCallbacks.size = 0;


	if (config.visuals.viewModel) {
		view_x->setValue(config.visuals.viewModel_x);
		view_y->setValue(config.visuals.viewModel_y);
		view_z->setValue(config.visuals.viewModel_z);
	}
}
```
Visuals.h in namespace Visuals() at the beginning.
```C++
void viewModel() noexcept;
```
GUI.cpp after hitmarker time (Remove XorStr ( ) if u dont have xorstr encryption in ur osiris)
```C++
		ImGui::Checkbox(XorStr("XYZ/BOB"), &config.visuals.viewModel);
		if (config.visuals.viewModel) {
			ImGui::PushID(6);
			ImGui::SliderFloat("", &config.visuals.viewModel_x, -10, 10, XorStr("ViewX:%f"));
			ImGui::PopID();
			ImGui::PushID(7);
			ImGui::SliderFloat("", &config.visuals.viewModel_y, -10, 10, XorStr("ViewY:%f"));
			ImGui::PopID();
			ImGui::PushID(8);
			ImGui::SliderFloat("", &config.visuals.viewModel_z, -10, 10, XorStr("ViewZ:%f"));
			ImGui::PopID();
		}
```
Config.h in visuals struct after hitmarkertime variable
```C++
		bool viewModel{ false };
		float viewModel_x{ 0 };
		float viewModel_y{ 0 };
		float viewModel_z{ 0 };
```
Config.cpp in visualsJson.isMember
```C++
[...]

PR #584, opened 2019-09-26
Added Custom View model Offset (Highly Requested)

effex1337:
I tried an unique approach for saving fps as it constantly changes value but if i keep a button it does its job but its a bit annoying. @danielkrupinski Please let me know your opinion.

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -364,6 +364,9 @@ void Config::load(size_t id) noexcept
         if (miscJson.isMember("Choked packets key")) misc.chokedPacketsKey = miscJson["Choked packets key"].asInt();
         if (miscJson.isMember("Grenade predict")) misc.nadePredict = miscJson["Grenade predict"].asBool();
         if (miscJson.isMember("Max angle delta")) misc.maxAngleDelta = miscJson["Max angle delta"].asFloat();
+        if (miscJson.isMember("Custom Viewmodel X")) misc.viewmodel_x = miscJson["Custom Viewmodel X"].asFloat();
+        if (miscJson.isMember("Custom Viewmodel Y")) misc.viewmodel_y = miscJson["Custom Viewmodel Y"].asFloat();
+        if (miscJson.isMember("Custom Viewmodel Z")) misc.viewmodel_z = miscJson["Custom Viewmodel Z"].asFloat();
     }
 
     {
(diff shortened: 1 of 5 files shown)


deprale:
Why not add a button like I did? :) Also static ConVar* is not needed.
You can just do the default static auto ones because theres a overrideview hook so you don't have to use the sv_competitive_minspec meme...
#574 check my reply :P

effex1337:
It does its job so yeah and did you mean checkbox by `Why not add a button like I did? :) `

effex1337:
and also i did with memory usage in mind.

effex1337:
@danielkrupunski can u please take a look?

deprale:
yeah i meant a checkbox cuz then every single time u inject u will have x y z offset all set to 0 instantly which is gay.
at least if u inject and have a checkbox u dont have this slight annoyance. 
also could be better if you just made a struct with saved original x y z structs and if you disable the checkbox just return the original value to the cvars including the callbacks :P but yea thats a bit more work and generally not needed for most people

xAkiraMiura:
how about you check if any of the values have changed across function calls and run it if they have? that would solve the issue of low fps due to constantly updating a cvar but also not require any user intervention.

DerLi1:
How can i do auto update?

xAkiraMiura:
you need to understand basic logic first. 

if (current == previous) // it changed
previous = current

RyDeem:
> you need to understand basic logic first.
> 
> if (current != previous) // it changed
> previous = current

Sorry for necro or whatever but

is this code better? 

`////
void Visuals::ViewmodelXYZ() noexcept {

    if (!localPlayer)
        return;

    int viewmodel_x = 0; int viewmodel_y = 0; int viewmodel_z = 0; int minspec = 0;
    interfaces->cvar->findVar("viewmodel_offset_x")->setValue(viewmodel_x);
    interfaces->cvar->findVar("viewmodel_offset_y")->setValue(viewmodel_y);
    interfaces->cvar->findVar("viewmodel_offset_z")->setValue(viewmodel_z);
    interfaces->cvar->findVar("sv_competitive_minspec")->setValue(minspec);

    if (minspec != 0)
        interfaces->cvar->findVar("sv_competitive_minspec")->setValue(0);
    if (viewmodel_x != config->visuals.viewmodel_x)
        interfaces->cvar->findVar("viewmodel_offset_x")->setValue(config->visuals.viewmodel_x);
    if (viewmodel_y != config->visuals.viewmodel_y)
        interfaces->cvar->findVar("viewmodel_offset_y")->setValue(config->visuals.viewmodel_y);
    if (viewmodel_z != config->visuals.viewmodel_z)
        interfaces->cvar->findVar("viewmodel_offset_z")->setValue(config->visuals.viewmodel_z);`

minspec is to move the value more then 2.5 each direction

// i dont master the code quote function in this forum xd

littlehurensohn:
Do null callbacks and everything be fine, like that u can risk untrusted.

xAkiraMiura:
> null callbacks 

do you know what a callback is?
C++ doesn't implement null and it is only ever defined in Windows.h as 0
you ment "check for nullptr"

PR #643, opened 2019-10-04
Fixed headDot

anonymouschicken111:
Here the headDot changes in size accordingly to the distance of the enemy so that the dot wouldn't become a nuisance, because enemies further back have really big head dot and enemies closer front have a small head dot. I have also change the headDot into a single circle.

### DIFF
--- a/Osiris/Hacks/Esp.cpp
+++ b/Osiris/Hacks/Esp.cpp
@@ -150,14 +150,30 @@ static void renderBox(Entity* entity, const decltype(config.esp[0])& config) noe
     }
 }
 
+float distance(Vector src, Vector des)
+{
+	const float distance = sqrt((src.x - des.x) * (src.x - des.x) + (src.y - des.y) * (src.y - des.y) + (src.z - des.z) * (src.z - des.z));
+	return distance * 0.0254f;
+}
+
 static constexpr void renderHeadDot(Entity* entity, const decltype(config.esp[0])& config) noexcept
 {
     if (config.headDot) {
         Vector head{ };
         if (worldToScreen(entity->getBonePosition(8), head)) {
             interfaces.surface->setDrawColor(config.headDotColor, 255);
-            for (int i = 1; i <= 3; i++)
-                interfaces.surface->drawOutlinedCircle(head.x, head.y, i, 100);
+            const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
+
+            Vector localPos{ localPlayer->origin() };
+            Vector entityPos{ entity->origin() };
+
+            float distanceToEntity{ distance(localPos, entityPos) };
+
+            int rad{ (int) ceil(39.0f / distanceToEntity) };
+
+            //int seg{ (int) ceil(rad * 35.0f) };
+
+            interfaces.surface->drawOutlinedCircle(head.x + 0.5f, head.y - 0.5f, rad, 20);
         }
     }
 }


anonymouschicken111:
The radius is inversely proportional to the distance of entity
`y = k / x`
I found that 13.f meters has the perfect radius of 3
So,
```
y = radius, x = distance
3 = k / 13
k = 39
```

distance() is basically Pythagoras theorem

tirziz:
Too much mathematics for me

anonymouschicken111:
lol, I'm 12 and I just learned Pythagoras theorem in school
Since I changed the filled circle into a circle, and if you insist on making it a filled circle here you go:

```
static constexpr void renderHeadDot(Entity* entity, const decltype(config.esp[0])& config) noexcept
{
	if (config.headDot)
	{
		Vector head{ };
		if (worldToScreen(entity->getBonePosition(8), head))
		{
			const auto localPlayer{ interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer()) };

			interfaces.surface->setDrawColor(config.headDotColor, 255);

			Vector localPos{ localPlayer->origin() };
			Vector entityPos{ entity->origin() };

			float distanceToEntity{ distance(localPos, entityPos) };

			int rad{ (int)ceil(39.0f / distanceToEntity) };

			int seg{ (int) ceil(rad * 35.0f) };

			for (int i = 1; i <= rad; i++)
				interfaces.surface->drawOutlinedCircle(head.x, head.y, i, seg);
		}
	}
}
```

Bywnb:
Thank you for pull requests

Osiris/Hacks/Esp.cpp
```diff
@@ -150,14 +150,30 @@ static void renderBox(Entity* entity, const decltype(config.esp[0])& config) noe
     }
 }
 
+float distance(Vector src, Vector des)
```
danielkrupinski: You should use Vector::length() instead.

Osiris/Hacks/Esp.cpp
```diff
@@ -150,14 +150,30 @@ static void renderBox(Entity* entity, const decltype(config.esp[0])& config) noe
...
+            Vector localPos{ localPlayer->origin() };
+            Vector entityPos{ entity->origin() };
+
+            float distanceToEntity{ distance(localPos, entityPos) };
+
+            int rad{ (int) ceil(39.0f / distanceToEntity) };
+
+            //int seg{ (int) ceil(rad * 35.0f) };
+
+            interfaces.surface->drawOutlinedCircle(head.x + 0.5f, head.y - 0.5f, rad, 20);
```
danielkrupinski: Use new circle drawing function -> https://github.com/danielkrupinski/Osiris/blob/add8abee259538ac1ee08c696542fe6ab04cdf71/Osiris/SDK/Surface.h#L105 which gives better results

mdilai: Hi Daniel, it's a _slightly_ offtopic but can you explain usage of new function? There is no comments or so and it not used anywhere in code itself. What is `int startRadius` and `int radius`? It's full analog of `int r` and `int seg` of drawOutlinedCircle() or draw logic must be re-implemented in some way?

danielkrupinski: @mdilai The purpose `startRadius` is to allow to draw rings / circle outline as well as full circle e.g startRadius = 0 and radius = 5 would create full circle, while startRadius = 4 and radius = 5 would create 1px-width ring

mdilai: Got it, thanks!

anonymouschicken111:
Thanks :)

PR #644, opened 2019-10-04
Increase Prepare Revolver Trigger Time

hyperxpro:
Increase Prepare Revolver Trigger Time from 0.234375f to 0.242f.


anonymouschicken111:
why thou?

hyperxpro:
> why thou?

Create a little long stretch sound of trigger pull during prepare and short duration of the fire after the trigger is pulled.

VuBrian22:
It's a small change, but it'll help well slightly. (And thank you for putting this in the right place!)

PR #743, opened 2019-11-04
Improve Color World & +Change Alpha Of World

JDeu:
The color world at the moment isn't very good.

This color world is more like the one you see in Borings video's and can change the alpha of the world textures (X-ray).

Have been a bit lazy and not conformed to the newest version but its only the config that needs to be changed. I also removed the rainbow effect as the color can only be set once, not constantly, otherwise it would reduce your FPS.

I think I have added everything from my local version, if not comment the issues.

Screenshot:
image

### DIFF
--- a/Osiris/Hooks.cpp
+++ b/Osiris/Hooks.cpp
@@ -168,7 +168,6 @@ static int __stdcall doPostScreenEffects(int param) noexcept
         Visuals::thirdperson();
         Misc::inverseRagdollGravity();
         Visuals::disablePostProcessing();
-        Visuals::colorWorld();
         Visuals::reduceFlashEffect();
         Visuals::removeBlur();
         Visuals::updateBrightness();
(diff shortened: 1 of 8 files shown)


thomas0aotian:
gj bud keep the work up. love it

Novalaki:
Nice work man! but you forget to add what you have for color world in GUI.cpp

Osiris/Hooks.cpp
```diff
@@ -193,6 +192,7 @@ static float __stdcall getViewModelFov() noexcept
 static void __stdcall drawModelExecute(void* ctx, void* state, const ModelRenderInfo& info, matrix3x4* customBoneToWorld) noexcept
 {
     if (interfaces.engine->isInGame()) {
+        Visuals::colorWorld();
```
danielkrupinski: You invoke Visuals::colorWorld() every time drawModelExecute hook is called. Call it only once per frame.

nothanku:
can you add a checkbox for modulating static props? so it modulates the props on the map with a darker color

danielkrupinski:
Thank you very much for submitting pull request! I've decided to implement your solution with rainbow effect and additional check to modify only precached materials.

JDeu:
Great 😄 

Also, using the "r_DrawSpecificStaticProp" cvar and the material name "StaticProp", you can also modulate Props like boxes.

PR #802, opened 2019-11-26
Added spam if you shoot an enemy

tothzalan:


effex1337:
Spray with an smg get kicked, best feature ever.

flucative:
@danielkrupinski Hey dude, could u please add a knifebot to the cheat? And improve the aimbot aswell since its not perfect

danielkrupinski:
I don't think it's useful.

PR #821, opened 2019-12-06
Dead ESP

ZerGo0:
Enables the ESP when you're dead. ESP itself doesn't need to be enabled obviously.


Shaxzy:
Awesome! Thank you

agshaidun:
🤩🤩🤩

danielkrupinski:
Is it supposed to apply only to Player ESP? If so you should move `deadesp` field from Shared to the Player struct.

ZerGo0:
> Is it supposed to apply only to Player ESP? If so you should move `deadesp` field from Shared to the Player struct.

Done, didn't even notice that. ^^

danielkrupinski:
Thanks!

PR #866, opened 2019-12-15
No zoom on scope

Khajiitos:
Really simple yet nice feature to have, might be useful for some people, what it does is there is no zoom on scope.


agshaidun:
Good for raging :D

PR #892, opened 2019-12-22
Add chat spam

ClaudiuHKS:
Add the chat spam feature


Shaxzy:
Thank you! Great 💃

xAkiraMiura:
Spam delay is unclear in the units (I assume ticks)
0 spam delay is a 1 way ticket to getting kicked for too many commands (maybe raise the minimum)
Wierd naming scheme (theTime instead of currTime)
Using system time instead of engine currTime, this means Osiris chat spammers will not sync up together.

^ i dont know how to do code reviews on github

ClaudiuHKS:
I really appreciate your help in giving ideas but if you don't help with the code, I'm afraid I can't do it as you suggest (so the cheaters can spam at the same time and stuff like that).

Sorry about that.

lgv-0:
The skinchanger change seems unrelated to the pull request.. But the way you're storing the "phrases" isn't very standard. While ImGUI supports [multi-line text boxes](https://github.com/ocornut/imgui/blob/bdd31ec4dbb6ec415168722392c4dae91e27b7d0/imgui_widgets.cpp#L3116-L3119), that's specifically made for multi-line text fields, which you can actually use in [other source games like TF2](https://github.com/nullworks/cathook/blob/23c46143f6620e37da79d3ed638d7adea76b825b/src/hacks/Spam.cpp#L385-L392), so this format wouldn't be viable for exporting. Plus, it isn't quite what that widget is meant for. JSON offers [Arrays](https://sites.cs.ucsb.edu/~pconrad/github/jsoncpp/dist/doxygen/jsoncpp-api-html-0.6.0-dev/class_json_1_1_value.html#ac8c898f93543e55b67418f94bced20af) for situations just like this, [where you have a list of something you want to store](https://nlohmann.github.io/json/classnlohmann_1_1basic__json_a8a7d537fc297bbc5b845f92fe0445e3b.html). But, this will pop up a new problem of what you would replace your text box with. A promising option is the [Listbox](https://github.com/ocornut/imgui/blob/b029182a737929f8ee81f41b02e06fcadfca7d95/imgui_demo.cpp#L599-L601), but I wouldn't leave it how it is.

ClaudiuHKS:
I am still waiting for @danielkrupinski to review this pull request and ask for changes, if needed.

It works perfectly for now and is not actually causing any issues or performance loss.

**std::string** is a very performant string class and it can store up to **8192** text phrases, each having up to **128** characters, separated by **\n** character.

As you see, **std::string::size_type** is a 32 bit unsigned type.

Issue #619, opened 2019-10-01
{suggestion} doorspam like in skeet


ccsimplyspolit:
add doorspam feature
doorspam.cpp
```#include "doorspam.h"

#include "../settings.h"
#include "../interfaces.h"

bool Settings::DoorSpam::enabled = false;
static bool ShouldSpamDoors = false;

void DoorSpam::CreateMove(CUserCmd* cmd)
{
	if (!Settings::DoorSpam::enabled)
		return;

	C_BasePlayer* localplayer = (C_BasePlayer*) entityList->GetClientEntity(engine->GetLocalPlayer());
	if (!localplayer)
		return;

	if (cmd->buttons & IN_USE){	

		for (int i = 1; i < entityList->GetHighestEntityIndex(); i++)
		{			
			C_BaseEntity* entity = entityList->GetClientEntity(i);
			if (!entity)
				continue;

			ClientClass* client = entity->GetClientClass();

			if (client->m_ClassID == EClassIds::CPropDoorRotating){
				if (localplayer->GetVecOrigin().DistTo(entity->GetVecOrigin()) < 100){
					if (ShouldSpamDoors){
						cmd->buttons &= ~IN_USE;
					} else {
						cmd->buttons |= IN_USE;				
					}
					ShouldSpamDoors = !ShouldSpamDoors;	
				}
			}
		}

	}
}
```

doorspam.h
```
#pragma once

#include "../SDK/IInputSystem.h"

namespace DoorSpam
{
	//Hooks
	void CreateMove(CUserCmd* cmd);
}
```

imgui
```
ImGui::Checkbox(XORSTR("Door Spam"), &Settings::DoorSpam::enabled);
```

anonymouschicken111:
You forgot Hooks.cpp and Config.h and Config.cpp
It's not ImGui it's GUI.cpp
It's not `CUserCmd*` cmd it's `UserCmd* cmd` (btw you can drop the cmd in the header file)
Osiris ain't got em' classes
`C_BaseEntity`
`ClientClass`

The config system's different
configs.doorSpam.enabled
Since you said door spam you have to add a struct named doorSpam in Config.h with a
`bool enabled{ true };`

In the head file it's
`#include "../SDK/InputSystem.h"`

Then you have to add it to Hooks.cpp

And also how the hell is doorSpam useful, I'm pretty sure Osiris is a legit cheat and you'd probably get vacced when use this cheat without junkcode and shit.

[...]

PR #904, opened 2019-12-24
Use spam

notgoodusename:
Spams e basically

### DIFF
--- a/Osiris/Hacks/Misc.h
+++ b/Osiris/Hacks/Misc.h
@@ -32,7 +32,7 @@ namespace Misc {
     void quickHealthshot(UserCmd*) noexcept;
     void fixTabletSignal() noexcept;
     void fakePrime() noexcept;
-
+    static void usespam(UserCmd*) noexcept;
     constexpr void fixMovement(UserCmd* cmd, float yaw) noexcept
     {
         if (config.misc.fixMovement) {
--- a/Osiris/Hacks/Misc.h
+++ b/Osiris/Hacks/Misc.h
@@ -32,7 +32,7 @@ namespace Misc {
     void quickHealthshot(UserCmd*) noexcept;
     void fixTabletSignal() noexcept;
     void fakePrime() noexcept;
-    static void usespam(UserCmd*) noexcept;
+    void usespam(UserCmd*) noexcept;
     constexpr void fixMovement(UserCmd* cmd, float yaw) noexcept
     {
         if (config.misc.fixMovement) {
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -1526,6 +1526,7 @@ void Config::save(size_t id) const noexcept
         miscJson["Animated clan tag"] = misc.animatedClanTag;
         miscJson["Fast duck"] = misc.fastDuck;
         miscJson["Moonwalk"] = misc.moonwalk;
+        miscJson["Use spam"] = misc.usespam;
         miscJson["Sniper crosshair"] = misc.sniperCrosshair;
         miscJson["Recoil crosshair"] = misc.recoilCrosshair;
         miscJson["Auto pistol"] = misc.autoPistol;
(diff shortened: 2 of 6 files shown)


anarbmw:
I thought u quitted hacking

@notgoodusename

notgoodusename:
> I thought u quitted hacking
> 
> @notgoodusename

idk what to say i just like coming back to the cheating scene

tirziz:
why this feature exist ?

notgoodusename:
#619

rosesware:
Why a static function?

notgoodusename:
i don remember why i used the static function, but i think it should work the same without it

danielkrupinski:
What is the purpose of this?

notgoodusename:
i use it to grab people gun when they throw it or troll people with doors with doors

danielkrupinski:
Wouldn't it make bomb undefusable for you?

notgoodusename:
yeah didnt think of that maybe i can check if the bomb is planted and disable it

notgoodusename:
that should fix it

Osiris/Hacks/Misc.h
```diff
@@ -32,7 +32,7 @@ namespace Misc {
     void quickHealthshot(UserCmd*) noexcept;
     void fixTabletSignal() noexcept;
     void fakePrime() noexcept;
-
+    static void usespam(UserCmd*) noexcept;
```
ClaudiuHKS: **static** is not good here
**static** is good here, for example, because it's only **Netvars.cpp** related: https://github.com/danielkrupinski/Osiris/blob/master/Osiris/Netvars.cpp#L11

rosesware:
> that should fix it

@notgoodusename More broken, even worse.

notgoodusename:
what do you mean "even worse"?

rosesware:
> what do you mean "even worse"?

The "bomb fix" don't work due to "many command's" and get kicked. 

> "even worse"

Because before you fix it, don't have this "many command's" error and get kicked from server.

notgoodusename:
> > what do you mean "even worse"?
> 
> The "bomb fix" don't work due to "many command's" and get kicked (weird).
> 
> > "even worse"
> 
> Because before you fix it, don't have this "many command's" error and get kicked from server.

Weird because it only sends actions not commands

rosesware:
> > > what do you mean "even worse"?
> > 
> > 
> > The "bomb fix" don't work due to "many command's" and get kicked (weird).
> > > "even worse"
> > 
> > 
> > Because before you fix it, don't have this "many command's" error and get kicked from server.
> 
> Weird because it only sends actions not commands

When I try to see if really get fixed on a localhost server with Bot's, I plant the bomb with "E" button and I get kicked by "To many command's" (I don't even test if works on matchmaking or something).

notgoodusename:
ohhh that makes sense

rosesware:
By the way, nice work with new Anti Aim improvement. I already had, 'cause the request before, now just needed to change and fix some thinks and is working fine, thanks.

notgoodusename:
> By the way, nice work with new Anti Aim improvement. I already had, 'cause the request before, now just needed to change and fix some thinks and is working fine, thanks.

the fix to usespam is this
 void Misc::usespam(UserCmd* cmd) noexcept
 {
     static bool usespam = true;
     const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
     if (localPlayer->isDefusing()) {
         return;
     }
     if (!localPlayer->isAlive() || localPlayer->inBombZone()){
         return;
 }
     if (config.misc.usespam && cmd->buttons & UserCmd::IN_USE) {
         if (usespam)
         {
             cmd->buttons |= UserCmd::IN_USE;
             usespam = false;
         }
         else
         {
             cmd->buttons &= ~UserCmd::IN_USE;
             usespam = true;
         }
     }
 }

notgoodusename:
> By the way, nice work with new Anti Aim improvement. I already had, 'cause the request before, now just needed to change and fix some thinks and is working fine, thanks.

Thanks!

PR #934, opened 2019-12-30
Team based player models

MinecraftGoodGame:
Ability to have different player models on different teams. Simple as that.


rosesware:
Interesting, but at the same time unnecessary.
Well, nice work. 👍

Straightplus:
maybe for skinchanger?

Straightplus:
Pretty good find :D

anonymouschicken1111:
@roseware might be useless to you but maybe not for others

danielkrupinski:
Thank you! I'd be great to also set model for player ragdoll (body falling on ground) as it reverts to default model.

MinecraftGoodGame:
I don't think there is a way to determine if ragdoll is localplayer, correct me if I'm wrong

Straightplus:
I aproove i actually like this thanks for sharing :D

PR #963, opened 2020-01-09
Autostrafe fix?

daisuki333:
I believe some values for the autostrafe were not right, and i feel this change fixes autostrafe cause it was kind of wonky before

### DIFF
--- a/Osiris/Hacks/Misc.h
+++ b/Osiris/Hacks/Misc.h
@@ -109,9 +109,9 @@ namespace Misc
             config.misc.autoStrafe
             && !(localPlayer->flags() & 1)
             && localPlayer->moveType() != MoveType::NOCLIP) {
-            if (cmd->mousedx < -20)
+            if (cmd->mousedx < 0)
                 cmd->sidemove = -450.0f;
-            else if (cmd->mousedx > 20)
+            else if (cmd->mousedx > 0)
                 cmd->sidemove = 450.0f;
         }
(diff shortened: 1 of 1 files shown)


ClaudiuHKS:
Thanks!

xAkiraMiura:
i think the original purpose of the delta check was to not make the autostrafe too smooth and have VACnet determine it was a robot rather than somebody pressing the keys themself, if you havent encountered any vac auths or bans while using the change then the original purpose is probably obsolete.

daisuki333:
> i think the original purpose of the delta check was to not make the autostrafe too smooth and have VACnet determine it was a robot rather than somebody pressing the keys themself, if you havent encountered any vac auths or bans while using the change then the original purpose is probably obsolete.

I mean i'm kind of new to c++ in a way which kind of makes me question if i'm right or wrong, but like i haven't necessarily gotten any vac auth errors, if i am wrong and people do get then, i apologise.

notgoodusename:
i dont think you can get vac auth from a legit autostrafe, like vac auth comes when you flick alot on casual,deathmatch,etc

xAkiraMiura:
valve probably isn't smart enough to check if your velocity always follows your view angle but if they are the slightest bit competent then it could become a detection vector in the future.

danielkrupinski:
It was done to not strafe while jumping on box but it seems velocity check would be better.

PR #1007, opened 2020-01-22
Provide hotkey for Auto Strafe

simonsmh:
Tested with edge jump with mouse key.


RyDeem:
i'd make the key function toggle, so people who enjoy allways on autostrafe can have it the old way, and who want it like this only hotkey can have it the new way.

I for one dont use autostrafe, but if id use it it would be on my bhop key.
 so I like hotkey autostrafe, but some people would probably like it on "toggle no key" like it was before, so its not wise imho to just "change" functionality to only work on hotkey as i see in the code, maybe i got this wrong.

would love to see some more input from other people.

one love peace man

simonsmh:
@RyDeem It should remain always on when users didn't set hotkey now.

Straightplus:
tanks men weryyy gud

PR #1015, opened 2020-01-24
Default Hitmarker

ZerGo0:
I could make it so that it fades out, but I'm too lazy right now.

Donated btw. I forgot to add my name to the message. ;)

### DIFF
--- a/Osiris/Hacks/Visuals.cpp
+++ b/Osiris/Hacks/Visuals.cpp
@@ -318,8 +318,7 @@ void Visuals::hitMarker(GameEvent* event) noexcept
 
     static float hitMarkerLastHitTime = 0.0f;
 
-    if (event && interfaces.engine->getPlayerForUserID(event->getInt("attacker")) == interfaces
-        .engine->getLocalPlayer())
+    if (event && interfaces.engine->getPlayerForUserID(event->getInt("attacker")) == localPlayer->index())
     {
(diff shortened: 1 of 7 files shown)


Osiris/Hacks/Visuals.cpp
```diff
@@ -307,3 +308,41 @@ void Visuals::hitEffect(GameEvent* event) noexcept
         }
     }
 }
+
+void Visuals::hitMarker(GameEvent* event) noexcept
+{
+    const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
+
+    if (!localPlayer || !localPlayer->isAlive() || !config.visuals.hitMarker || config.visuals.hitMarker == 0)
```
ClaudiuHKS: `if (!localPlayer || !localPlayer->isAlive() || !config.visuals.hitMarker || config.visuals.hitMarker == 0)`

→

`if (!localPlayer || !localPlayer->isAlive() || config.visuals.hitMarker == 0)`

ZerGo0: 🤫 fixed

notgoodusename:
instead of switch do if(int != 0)

ZerGo0:
People might want to add more variations in the future, that's why I coded it like that.

danielkrupinski:
Thanks for your contribution! Donation appreciated 😊

PR #1003, opened 2020-01-20
Added bomb damage indicator

Laxybaxy:
predicts bomb damage based on armor value and position from bomb.

its positioned in bottom right corner, kinda hard to spot but ye

video: https://youtu.be/8nHfuEH7840

Thanks to Casual_Hacker from UC

PR #1016, opened 2020-01-24
Bomb Damage Indicator

ZerGo0:
Proper Bomb Damage Indicator with GUI checkbox, Config entries and at the right location on the screen. 

Thanks to @Laxybaxy for providing the actual code in #1003 

Preview:
pre1
pre2

It loops through the entity-list as well which isn't really optimal, I could remove the checkbox and shit and just close this and request Laxybaxy to change the position. Tell me what you think.


RyDeem:
@Laxybaxy had gui and config and stuff, i just suggested he integrates the damage function to the bomb timer info panel, as if to merge it into an bomb info panel.

So this improves on the work by getting the text in the same position as bomb timer and shows for spectated entity as well? If so its nice to say to a teammate to run further away if they gon die n ye.

RyDeem:
if that would be done the function could be renamed to bomb info not bomb timer

Osiris/Hacks/Misc.cpp
```diff
@@ -506,3 +506,64 @@ void Misc::killMessage(GameEvent& event) noexcept
...
+
+    for (int i = interfaces.engine->getMaxClients(); i <= interfaces.entityList->getHighestEntityIndex(); i++)
+    {
+        auto entity = interfaces.entityList->getEntity(i);
+        if (!entity || entity->isDormant() || entity->getClientClass()->classId != ClassId::PlantedC4 || !entity->c4Ticking())
+            continue;
+
+        auto vecBombDistance = entity->origin() - localPlayer->origin();
+
+        const auto a = 450.7f;
```
xAkiraMiura: magic numbers.
excessive use of auto.
a, b, and c are only used once so why do they need to be a variable?

Osiris/Hacks/Misc.cpp
```diff
@@ -506,3 +506,64 @@ void Misc::killMessage(GameEvent& event) noexcept
...
+            auto flArmor = (flDamage - flNew) * flArmorBonus;
+
+            if (flArmor > static_cast<float>(ArmorValue)) {
+                flArmor = static_cast<float>(ArmorValue)* (1.f / flArmorBonus);
+                flNew = flDamage - flArmor;
+            }
+
+            flDamage = flNew;
+        }
+        const auto bombDamage = max((int)ceilf(flDamage), 0);
```
xAkiraMiura: inconsistance casting. C-style casting right here but in other places static_cast is used.
wouldn't flooring an int be faster than flooring a float?
^ convert to int then floor

Osiris/Hacks/Misc.cpp
ZerGo0: No idea ask the original author about these.

Osiris/Hacks/Misc.cpp
ZerGo0: Casting it to an int before flooring it would result in a wrong result since it will round down.

xAkiraMiura: i failed to consider the ceiling operation was dependant on the fact it was a float.
it should still be changed to a <static_cast> instead of a C style cast, better consistancy and is checked by the compiler.

Osiris/Hacks/Misc.cpp
Laxybaxy: no need, i was just lazy as my name may indicate

Osiris/Hacks/Misc.cpp
```diff
@@ -506,3 +506,62 @@ void Misc::killMessage(GameEvent& event) noexcept
     cmd += "\"";
     interfaces.engine->clientCmdUnrestricted(cmd.c_str());
 }
+
+void Misc::drawBombDamage() noexcept
+{
+    const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
```
Laxybaxy: no need to define localplayer if config.misc.bombDamage is false

danielkrupinski:
The problem is this is highly inaccurate for maps with nonstandard bomb radius, e.g. Studio.

ZerGo0:
> The problem is this is highly inaccurate for maps with nonstandard bomb radius, e.g. Studio.

Yea, even on maps like Dust2 it's not that accurate sadly, I still have it in my build, but I don't fully trust it.

xAkiraMiura:
isn't the bomb damage/radius a netvar?
if it's not then you could parse the BSP file, slow but 100% reliable.

danielkrupinski:
@xAkiraMiura I'm afraid it's not. Bomb damage is calculated entirely on the server.

Osiris/Hacks/Misc.cpp
ZerGo0: Fixed.

steel4me:
Thank you! Works great!

Issue #1036, opened 2020-01-30
game update


Bywnb:
i know GetCSWeaponData is 459
has any other index changed?

CerberusGame:
GetInaccuracy() 481
GetSpread() 451
UpdateAccuracyPenalty() 482
GetCSWpnData() 459
GetName() 385

Bywnb:
> GetInaccuracy() 481
> GetSpread() 451
> UpdateAccuracyPenalty() 482
> GetCSWpnData() 459
> GetName() 385

i update osiris including GetInaccuracy and GetCSWpnData , but still crush.Can u help me plz.

EurobeatClub:
Update isn't crashing me but aimbot did stop working.

Bywnb:
> Update isn't crashing me but aimbot did stop working.

yep, aimbot not works

ZerGo0:
Just attach a debugger and see which value doesn't update/return anything. Waiting for someone to fix it would probably also work.

ifoverload:
> GetInaccuracy() 481
> GetSpread() 451
> UpdateAccuracyPenalty() 482
> GetCSWpnData() 459
> GetName() 385

to me it does crashing

simonsmh:
```
struct WeaponData {
    std::byte pad1[4];   //0x0000
    char *name;          //0x0004
    std::byte pad1_[12]; //0x0008
    int maxClip;         //0x0014
    std::byte pad2[180]; //0x0018
    int type;            //0x00CC
    std::byte pad2_[28]; //0x00D0
    bool fullAuto;       //0x00EC
    std::byte pad3[3];   //0x00ED
    int damage;          //0x00F0
    float armorRatio;    //0x00F4
    int bullets;         //0x00F8
    float penetration;   //0x00FC
    std::byte pad4[8];   //0x0100
    float range;         //0x0108
    float rangeModifier; //0x010C
    std::byte pad5[16];  //0x0110
    bool silencer;       //0x0120
    std::byte pad6[15];  //0x0121
    float maxSpeed;      //0x0130
    float maxSpeedAlt;   //0x0134
};
```
https://github.com/simonsmh/Osiris/commit/9d730c5fa956b4f7bc84e25fa3e23ebeaa622efb

h4x0reL:
Just change this two bad boys and work good.
GetInaccuracy() 481
GetWeaponData() 459

Asviix:
[...]

PR #1037, opened 2020-01-30
Update offsets

simonsmh:
#1036

### DIFF
--- a/Osiris/SDK/Entity.h
+++ b/Osiris/SDK/Entity.h
@@ -164,17 +164,17 @@ public:
 
     constexpr int getWeaponSubType() noexcept
     {
-        return callVirtualMethod<int>(this, 279);
+        return callVirtualMethod<int>(this, 281);
     }
 
     constexpr WeaponData* getWeaponData() noexcept
     {
-        return callVirtualMethod<WeaponData*>(this, 457);
+        return callVirtualMethod<WeaponData*>(this, 459);
     }
 
     constexpr float getInaccuracy() noexcept
     {
-        return callVirtualMethod<float>(this, 479);
+        return callVirtualMethod<float>(this, 481);
     }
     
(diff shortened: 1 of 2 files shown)


ClaudiuHKS:
Awesome work, thanks!

tamas0313:
Névtelen
Esp Active Weapon isn't working!

simonsmh:
Might be getWeaponData problem. let me debug for a while.

rosesware:
And "Auto wall" is not working.

simonsmh:
@tamas0313 
image
Well I think there should be some problem in `interfaces.localize->find()` that it couldn't find translation? I think it is not related to this pull request.
I force print the name instead to test the result.
```
diff --git a/Osiris/Hacks/Esp.cpp b/Osiris/Hacks/Esp.cpp
index e8d589a8..f513a019 100644
--- a/Osiris/Hacks/Esp.cpp
+++ b/Osiris/Hacks/Esp.cpp
@@ -310,7 +310,10 @@ static void renderWeaponBox(Entity* entity, const Config::Esp::Weapon& config) n
         renderBox(bbox, config);

         if (config.name.enabled) {
-            const auto name{ interfaces.localize->find(entity->getWeaponData()->name) };
+            const size_t cSize = strlen(entity->getWeaponData()->name) + 1;
+            wchar_t *wc = new wchar_t[cSize];
+            mbstowcs(wc, entity->getWeaponData()->name, cSize);
+            const auto name{wc};
             const auto [width, height] { interfaces.surface->getTextSize(config.font, name) };
             interfaces.surface->setTextFont(config.font);
             if (config.name.rainbow)
```

notgoodusename:
this is how daniel does it every update 
https://github.com/danielkrupinski/Osiris/commit/d92e6e0b5039562d7482bea16a89e2b29658d7a8

PR #1040, opened 2020-01-30
Engine Prediction

notgoodusename:
All credits to @mdilai for making this!

### DIFF
--- a/Osiris/Hacks/EnginePrediction.cpp
+++ b/Osiris/Hacks/EnginePrediction.cpp
@@ -1,6 +1,6 @@
 #include "../SDK/GlobalVars.h"
 #include "../SDK/UserCmd.h"
-#include "EnginePrediction.h"
+#include "../Hacks/EnginePrediction.h"
 #include "../SDK/Prediction.h"
 #include "../Interfaces.h"
 #include "../SDK/PseudoMd5.h"
--- a/Osiris/Hacks/EnginePrediction.cpp
+++ b/Osiris/Hacks/EnginePrediction.cpp
@@ -1,6 +1,6 @@
 #include "../SDK/GlobalVars.h"
 #include "../SDK/UserCmd.h"
-#include "../SDK/EnginePrediction.h"
+#include "EnginePrediction.h"
 #include "../SDK/Prediction.h"
 #include "../Interfaces.h"
 #include "../SDK/PseudoMd5.h"
--- a/Osiris/Hacks/EnginePrediction.cpp
+++ b/Osiris/Hacks/EnginePrediction.cpp
@@ -1,0 +1,44 @@
+#include "../SDK/GlobalVars.h"
+#include "../SDK/UserCmd.h"
+#include "../SDK/EnginePrediction.h"
+#include "../SDK/Prediction.h"
+#include "../Interfaces.h"
+#include "../SDK/PseudoMd5.h"
+
+static float m_flOldCurtime;
+static float m_flOldFrametime;
+MoveData m_MoveData;
+
+void PredictionSys::RunEnginePred(UserCmd* cmd) noexcept
+{
+    const auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
+
+    *memory.predictionRandomSeed = MD5_PseudoRandom(cmd->commandNumber) & 0x7FFFFFFF;
+
+    m_flOldCurtime = memory.globalVars->currenttime;
+    m_flOldFrametime = memory.globalVars->frametime;
+
+    memory.globalVars->currenttime = memory.globalVars->serverTime(cmd);
+    memory.globalVars->frametime = memory.globalVars->intervalPerTick;
+
+    interfaces.gameMovement->StartTrackPredictionErrors(localPlayer);
+
+    memset(&m_MoveData, 0, sizeof(m_MoveData));
+    memory.moveHelper->SetHost(localPlayer);
+    interfaces.prediction->SetupMove(localPlayer, cmd, memory.moveHelper, &m_MoveData);
+    interfaces.gameMovement->ProcessMovement(localPlayer, &m_MoveData);
+    interfaces.prediction->FinishMove(localPlayer, cmd, &m_MoveData);
+}
+
+void PredictionSys::EndEnginePred() noexcept
+{
+    auto localPlayer = interfaces.entityList->getEntity(interfaces.engine->getLocalPlayer());
+
+    interfaces.gameMovement->FinishTrackPredictionErrors(localPlayer);
+    memory.moveHelper->SetHost(nullptr);
+
+    *memory.predictionRandomSeed = -1;
+
+    memory.globalVars->currenttime = m_flOldCurtime;
+    memory.globalVars->frametime = m_flOldFrametime;
+}
--- a/Osiris/Hacks/EnginePrediction.h
+++ b/Osiris/Hacks/EnginePrediction.h
@@ -1,0 +1,6 @@
+#pragma once
+
+	namespace PredictionSys {
+		void RunEnginePred(UserCmd*) noexcept;
+		void EndEnginePred() noexcept;
+	}
--- a/Osiris/Hooks.cpp
+++ b/Osiris/Hooks.cpp
@@ -24,6 +24,7 @@
 #include "Hacks/SkinChanger.h"
 #include "Hacks/Triggerbot.h"
 #include "Hacks/Visuals.h"
+#include "Hacks/EnginePrediction.h"
 
 #include "SDK/Engine.h"
 #include "SDK/Entity.h"
@@ -124,8 +125,6 @@ static bool __stdcall createMove(float inputSampleTime, UserCmd* cmd) noexcept
     Misc::bunnyHop(cmd);
     Misc::autoStrafe(cmd);
     Misc::removeCrouchCooldown(cmd);
-    Aimbot::run(cmd);
-    Triggerbot::run(cmd);
     Misc::autoPistol(cmd);
     Misc::autoReload(cmd);
     Misc::updateClanTag();
@@ -133,7 +132,6 @@ static bool __stdcall createMove(float inputSampleTime, UserCmd* cmd) noexcept
     Misc::fakeBan();
     Misc::stealNames();
     Misc::revealRanks(cmd);
-    Backtrack::run(cmd);
     Misc::quickReload(cmd);
     Misc::moonwalk(cmd);
     Misc::quickHealthshot(cmd);
@@ -145,7 +143,13 @@ static bool __stdcall createMove(float inputSampleTime, UserCmd* cmd) noexcept
         Misc::chokePackets(sendPacket);
         AntiAim::run(cmd, previousViewAngles, currentViewAngles, sendPacket);
     }
-
+    
+    PredictionSys::RunEnginePred(cmd);
+    Aimbot::run(cmd);
+    Triggerbot::run(cmd);
+    Backtrack::run(cmd);
+    PredictionSys::EndEnginePred();
+    
     auto viewAnglesDelta{ cmd->viewangles - previousViewAngles };
     viewAnglesDelta.normalize();
     viewAnglesDelta.x = std::clamp(viewAnglesDelta.x, -config.misc.maxAngleDelta, config.misc.maxAngleDelta);
--- a/Osiris/Interfaces.h
+++ b/Osiris/Interfaces.h
@@ -23,6 +23,8 @@ class ResourceAccessControl;
 class Surface;
 class Sound;
 class SoundEmitter;
+class Predict;
+class GameMovement;
 
 class Interfaces {
 public:
@@ -45,6 +47,8 @@ public:
     Surface* surface = find<Surface>(L"vguimatsurface", "VGUI_Surface031");
     Sound* sound = find<Sound>(L"engine", "IEngineSoundClient003");
     SoundEmitter* soundEmitter = find<SoundEmitter>(L"soundemittersystem", "VSoundEmitter003");
+    Predict* prediction = find<Predict>(L"client_panorama", "VClientPrediction001");
+    GameMovement* gameMovement = find<GameMovement>(L"client_panorama", "GameMovement001");
 private:
     template <typename T>
     static auto find(const wchar_t* module, const char* name) noexcept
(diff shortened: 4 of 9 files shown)


anarbmw:
@notgoodusename would u explain what is this?xd

Osiris/Hacks/EnginePrediction.cpp
```diff
@@ -0,0 +1,44 @@
+#include "../SDK/GlobalVars.h"
+#include "../SDK/UserCmd.h"
+#include "../SDK/EnginePrediction.h"
```
rosesware: `#include "../SDK/EnginePrediction.h"`
Changed to:
`#include "../Hacks/EnginePrediction.h"`
To fix.

rosesware:
> @notgoodusename would u explain what is this?xd

Like the name say, it's a engine who makes "Aimbot" predict and don't miss random shots (making "Aimbot" more accurate).

Osiris/Hacks/EnginePrediction.cpp
NexSqaud: to:  `#include "EnginePrediction.h"`

TheFantasticLoki:
Thanks I added it to my source and it seems to work great I already gave you credit in my readme.

mdilai:
@notgoodusename Oh dude, you brilliant and found my little treasure! During last year i did about 300 commits over my fork of Osiris, rewritten half of the cheat and become scared how to push all this things to mainline Osiris. Engine Prediction is only a drop in the sea, there a lot of changes, most notably for me is a fully re-implemented rage-capable AimBot with hitchance, velocity extrapolation and extreme optimization (about 15x time faster than original). 

@TheFantasticLoki in such case i will be glad if you add link to my original repo https://github.com/mdilai/JweegaCSGO . You can find a lot more things to paste there, just hope you are familiar with the debugger, my fork may be far away from stability and it semi-abandoned because i got bored from CSGO, cheating and freaking C++ xD

ZerGo0:
> > @notgoodusename would u explain what is this?xd
> 
> Like the name say, it's a engine who makes "Aimbot" predict and don't miss random shots (making "Aimbot" more accurate).

So it's something that you would only use to be more blatant, right? Are there any other use cases?

effex1337:
No @ZerGo0

effex1337:
In fact its gonna look better

mdilai:
> > > @notgoodusename would u explain what is this?xd
> > 
> > 
> > Like the name say, it's a engine who makes "Aimbot" predict and don't miss random shots (making "Aimbot" more accurate).
> 
> So it's something that you would only use to be more blatant, right? Are there any other use cases?

I will try to explain. Engine prediction simulation is a critical feature for any cheat. CSGO server and client exchange data 64/128 times per second (ticks). Every time when server receive you current position and action, it can propagate this data to client only in next tick, e.g. with 15ms delay. Imagine you looking a corner with AWP, suddenly got headshot and only 15ms later you will see the guy who killed you. Or trying to shoot running enemy, but every bullet will miss because enemy actually ran away from your crosshair 15ms ago.

To avoid this, CSGO doesn't propagate clients data directly. Instead, server use your velocity, speed, height etc. to calculate and predict where you will be in next 15ms if you will continue to run with same speed. And this data come to all players on server. So in "corner example" you will see enemy at same time when they will see you, even if server not yet received this data from enemy. This is called server-side Engine Prediction by Valve.

Osiris problem is that for math calcuations for Aimbot, Backtrack, Triggerbot (and maybe Anti-Aim, but not in current simple implementation) it using data directly from your client. So Aimbot got your crosshair location, got location of enemy head, calculate relative angle from crosshair to head and send it to server. But then server will "predict" your data, will set new position of you and enemy and now your calculated angle is wrong and point nowhere near the enemy as you has expected. If both you and enemy not moving there is no problem, but as soon as any of you start to moving Aimbot/Backtrack/Triggerbot have no chance to hit the enemy.
Thats why i've implemented simulation of server-side prediction in cheat, so first we will calculate how server will treat our position and then "feed" Aimbot with this data instead of original client data. That's how all perfectly match.
Most notably effect of engine prediction which easiest to see is RCS. Only with engine prediction you can get perfect RCS when every bullet is going to same point each time, because recoil are predicted too. Without it recoil control always few bullets away from real pattern.

I hope I clearly explained how this work. Ask your questions if any.

Osiris/Hacks/EnginePrediction.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include "EnginePrediction.h"
+#include "../SDK/Prediction.h"
+#include "../Interfaces.h"
+#include "../SDK/PseudoMd5.h"
+
+static float m_flOldCurtime;
+static float m_flOldFrametime;
+MoveData m_MoveData;
+
+void PredictionSys::RunEnginePred(UserCmd* cmd) noexcept
```
ClaudiuHKS: In my opinion, you should let the user enable or disable this feature.

``if (!config.misc.enginePrediction) return;``

Same for [ **void PredictionSys::EndEnginePred() noexcept** ].

You will need to edit **Config.h**, **Config.cpp** and **GUI.cpp** files.

notgoodusename: why??? we want engine prediction on all the time is a plus for the cheat

ClaudiuHKS: Maybe some people don't want it all the time. I just gave an idea.

notgoodusename: after reading mdlai explaination i think is best to have it on all the time

ZerGo0:
> > > > @notgoodusename would u explain what is this?xd
> > > 
> > > 
> > > Like the name say, it's a engine who makes "Aimbot" predict and don't miss random shots (making "Aimbot" more accurate).
> > 
> > 
> > So it's something that you would only use to be more blatant, right? Are there any other use cases?
> 
> I will try to explain. Engine prediction simulation is a critical feature for any cheat. CSGO server and client exchange data 64/128 times per second (ticks). Every time when server receive you current position and action, it can propagate this data to client only in next tick, e.g. with 15ms delay. Imagine you looking a corner with AWP, suddenly got headshot and only 15ms later you will see the guy who killed you. Or trying to shoot running enemy, but every bullet will miss because enemy actually ran away from your crosshair 15ms ago.
> 
> To avoid this, CSGO doesn't propagate clients data directly. Instead, server use your velocity, speed, height etc. to calculate and predict where you will be in next 15ms if you will continue to run with same speed. And this data come to all players on server. So in "corner example" you will see enemy at same time when they will see you, even if server not yet received this data from enemy. This is called server-side Engine Prediction by Valve.
> 
> Osiris problem is that for math calcuations for Aimbot, Backtrack, Triggerbot (and maybe Anti-Aim, but not in current simple implementation) it using data directly from your client. So Aimbot got your crosshair location, got location of enemy head, calculate relative angle from crosshair to head and send it to server. But then server will "predict" your data, will set new position of you and enemy and now your calculated angle is wrong and point nowhere near the enemy as you has expected. If both you and enemy not moving there is no problem, but as soon as any of you start to moving Aimbot/Backtrack/Triggerbot have no chance to hit the enemy.
> Thats why i've implemented simulation of server-side prediction in cheat, so first we will calculate how server will treat our position and then "feed" Aimbot with this data instead of original client data. That's how all perfectly match.
> Most notably effect of engine prediction which easiest to see is RCS. Only with engine prediction you can get perfect RCS when every bullet is going to same point each time, because recoil are predicted too. Without it recoil control always few bullets away from real pattern.
> 
> I hope I clearly explained how this work. Ask your questions if any.

Wow, thank you very much for the explanation, very appreciated.

anarbmw:
@notgoodusename does that mean that frickin triggerbot wont miss anymore?

notgoodusename:
nope

xAkiraMiura:
does engine prediction matter when you can just backtrack them?
I know you need it in HvH or you will be a tick back and you need it for yourself when doing movement stuff but is there any other use for it?

notgoodusename:
backtracking is a bad solution for hvh and with engine prediction you can do proper edgejump

mdilai:
> @notgoodusename does that mean that frickin triggerbot wont miss anymore?
Should help a bit when you use triggerbot to flickshot but not waiting moving target.
Both Aimbot and Triggerbot in Osiris actually are trash itself. I found enough time and enthusiasm only to rewrite Aimbot but not trigger. If somebody have enthusiasm, this lambda function will help you:
```
constexpr auto velocityExtrapolate {
    [](Entity* entity, const Vector& destination) constexpr noexcept {
        return destination + (entity->velocity() * memory.globalVars->intervalPerTick);
    }
};
```

Use it to extrapolate destination before sending to traceRay/lineGoesThroughSmoke/etc. Or paste my aimbot and use autoshot on button with low fov, actually same result but with all needed predictions and extrapolation. Maybe someday i will make a pull request...

Osiris/Hacks/EnginePrediction.cpp
mdilai: Actually there is no reason to disable prediction and you will not find any cheats which allow you to disable prediction. It's just useless.

notgoodusename:
> > @notgoodusename does that mean that frickin triggerbot wont miss anymore?
> 
> Should help a bit when you use triggerbot to flickshot but not waiting moving target.
> Both Aimbot and Triggerbot in Osiris actually are trash itself. I found enough time and enthusiasm only to rewrite Aimbot but not trigger. If somebody have enthusiasm, this lambda function will help you:
> 
> ```
> constexpr auto velocityExtrapolate {
>     [](Entity* entity, const Vector& destination) constexpr noexcept {
>         return destination + (entity->velocity() * memory.globalVars->intervalPerTick);
>     }
> };
> ```
> 
> Use it to extrapolate destination before sending to traceRay/lineGoesThroughSmoke/etc. Or paste my aimbot and use autoshot on button with low fov, actually same result but with all needed predictions and extrapolation. Maybe someday i will make a pull request...

what does velocityextrapolate do i saw it on your commit and you removed it i think

mdilai:
@notgoodusename I use it in Aimbot.cpp, just moved deeper near usage place. It's very dumb "prediction" of enemy, because proper engine prediction work only on local player. We get velocity of enemy per ms, multiply by interval per ticks (1/64s ~ 15ms) to get a distance which enemy pass in one tick and add it to known enemy position to get future enemy position in next tick (which actually will be interpreted by server). Real engine prediction using not only a velocity but also a height, jumps, angle of surface etc. but this is a minimum enough to shot at least a target moving on by a straight line.

atk3001:
I realy hope he adds it. Sadly this breaks the Desync AntiAim from you. If you turn on yaw with this in the cheat you will spin like crazy in first person.

notgoodusename:
> I realy hope he adds it. Sadly this breaks the Desync AntiAim from you. If you turn on yaw with this in the cheat you will spin like crazy in first person.

didnt have any problems with desync, what do u mean?

atk3001:
As soon as I turn AA on with yaw it seems like my first person synchronises with AA. Also I am nat walking in the right direction with AA on. 
But I made a Video of it. https://www.youtube.com/watch?v=L5ATk_0S7oY

notgoodusename:
enable fix movement

atk3001:
Ok that fixed the weird sideways walking but the spinning when turning on Aimbot is still there

notgoodusename:
please explain your problem

atk3001:
It looks like it only spins when standalone rcs is also on. So that seems to be the problem.

atk3001:
But it worked good before adding Engine Prediction.

notgoodusename:
like explain what do u mean?

atk3001:
You can see the spinning in the video at 20 seconds https://www.youtube.com/watch?v=L5ATk_0S7oY

notgoodusename:
i think the reason behind is that rcs i trying to compansate for the aa

atk3001:
But what do you do without rcs? I think you barely can use the cheat without the rcs.

notgoodusename:
i remade my aa (doesnt work but is better) and for me normal rcs works i havent tested standalone rcs with aa

Laxybaxy:
@mdilai @notgoodusename how would i get local flags from engineprediction?

notgoodusename:
the same way yo did before

mdilai:
> @mdilai @notgoodusename how would i get local flags from engineprediction?

You can’t. Local flags get corrupted due to forcing engine prediction in CreateMove hook. You need to do all local flags actions like bunnyhop before StartPrediction() or backup flags to global variable before calling startprediction and restore it after EndPrediction().

Laxybaxy:
yeye i am saving local flags before prediction, but i need to compare them to predicted flags inside the prediction, but how do i get these predicted flags`?

mdilai:
> yeye i am saving local flags before prediction, but i need to compare them to predicted flags inside the prediction, but how do i get these predicted flags`?

I am not really understand what you mean. Show example of code which give you issues.

notgoodusename:
you should do it after end prediction

notgoodusename:
to get predicted flags

mdilai:
> you should do it after end prediction

EndPrediction do not restore flags, mostly it just clear own internal buffers before next run. He should put his code **before StartPrediction**. 
By the way, antiaim should be after StartPrediction, not before.

notgoodusename:
i mean to compare unpredicted flags and predicted ones you have to call one func before startprediction and one after the prediction ends thats how i did edge jump

Laxybaxy:
Ok, gonna try that

Laxybaxy:
ok, so edgejump works now, but it works bad, it jumps when the origin of the player is not on the ground, any way to make it jump at last second?

this is the current one:
`    if (!(localPlayer->flags() & FL_ONGROUND) && (flagsBackup & FL_ONGROUND))
        cmd->buttons &= ~UserCmd::IN_JUMP;`

danielkrupinski:
What is the role of `predictionRandomSeed`?

notgoodusename:
> What is the role of `predictionRandomSeed`?

"PredictionRandomSeed is in fact used to handle spread values shortly after landing and at the beginning of your jump. This will obviously have an impact while playing on nospread servers. So leaving it out isn't necessarily a good thing if you want to truly do engine prediction correctly. As far as currenttime and currentcommand go, they don't get referenced anywhere after they have been set."

https://www.unknowncheats.me/forum/1546350-post2.html

notgoodusename:
sorry for answering late

streamshark:
the highly requested auto-jumpbug might become feasible with this, I had been trying to do it with basic velocity extrapolation and only got it to work on flat ground

Rabbe00:
Hello, when I try to compile this I get these errors. Any way to fix them?
image

streamshark:
I think I resolved that by adding engineprediction.cpp to the solution, it wasn't there for some reason for me

Straightplus:
cool

Rabbe00:
> I think I resolved that by adding engineprediction.cpp to the solution, it wasn't there for some reason for me

Still getting same errors :(

Laxybaxy:
Include enginepred.h aswell

streamshark:
looks like Daniel just started committing engine prediction today

notgoodusename:
> looks like Daniel just started committing engine prediction today

bruh he truly hates my prs xd

PR #1042, opened 2020-01-30
Dead thirdperson

JDeu:
As title says...

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -663,6 +663,7 @@ void Config::load(size_t id) noexcept
         if (visualsJson.isMember("Zoom")) visuals.zoom = visualsJson["Zoom"].asBool();
         if (visualsJson.isMember("Zoom key")) visuals.zoomKey = visualsJson["Zoom key"].asInt();
         if (visualsJson.isMember("thirdperson")) visuals.thirdperson = visualsJson["thirdperson"].asBool();
+        if (visualsJson.isMember("deadThirdperson")) visuals.deadThirdperson = visualsJson["deadThirdperson"].asBool();
         if (visualsJson.isMember("thirdpersonKey")) visuals.thirdpersonKey = visualsJson["thirdpersonKey"].asInt();
         if (visualsJson.isMember("thirdpersonDistance")) visuals.thirdpersonDistance = visualsJson["thirdpersonDistance"].asInt();
         if (visualsJson.isMember("viewmodelFov")) visuals.viewmodelFov = visualsJson["viewmodelFov"].asInt();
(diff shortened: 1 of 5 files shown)


xAkiraMiura:
Is this always on when dead or toggle-able like when alive?

JDeu:
It acts the same as the normal third person apart from you can’t set a distance. There is a slight issue that the player card stays on your screen while spectating, but I haven’t found a fix yet.

notgoodusename:
Where did you get the observermode class i need it for a better spec list

xAkiraMiura:
m_iObserverMode is a integer netvar.
you can use a netvar dumper to get it.

ObserverMode is an enum, not a class.

You could just use int and use `if (observerMode == 4) {}`, works fine, only issue is that they become magic numbers.

notgoodusename:
i want to do a switch like third person first person and free cam

TheFantasticLoki:
If you can get this fixed I'd love to have it added.

RyDeem:
> It acts the same as the normal third person apart from you can’t set a distance. There is a slight issue that the player card stays on your screen while spectating, but I haven’t found a fix yet.

so i got an idea, make the regular thirdperson distance apply when dead aswell here

effex1337:
If the current code for regular thirdperson distance doesn't work then you'll need to use a cvar for it 
cam_idealdist or smthin

rosesware:
It's broken.

JDeu:
> If the current code for regular thirdperson distance doesn't work then you'll need to use a cvar for it
> cam_idealdist or smthin

There is a netvar `m_flObserverChaseDistance` but the max distance is defined to 96.f so can’t really change it to most people’s third person distance.

playday3008:
now we have getObserverMode, sooo, how can I use Dead thirdperson with `localPlayer->getObserverMode()`
bcoz `localPlayer->getObserverMode() = (!config->visuals.thirdpersonKey || isInThirdperson) ? ObsMode::Chase : ObsMode::InEye;` have C2106 error

playday3008:
@JDeu

JDeu:
> now we have getObserverMode, sooo, how can I use Dead thirdperson with `localPlayer->getObserverMode()`
> bcoz `localPlayer->getObserverMode() = (!config->visuals.thirdpersonKey || isInThirdperson) ? ObsMode::Chase : ObsMode::InEye;` have C2106 error

GetObserverMode is a function that returns the enum, not sets. You need to add the netvar `m_iObserverMode`, which allows you to change the current observer mode.

playday3008:
@JDeu ok, and why `OBS_MODE_ROAMING = 7`? bcoz in getObserverMode, `Roaming = 6`
image

JDeu:
> @JDeu ok thanks, and why `OBS_MODE_ROAMING = 7`? bcoz in getObserverMode, `Roaming = 6`
> image

It is 6? the enum starts at index 0

PR #1129, opened 2020-02-20
Triggerbot hitchance

pubsense:
just 2 lines of code and now triggerbot stop spamming shots

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -72,6 +72,8 @@ void Config::load(size_t id) noexcept
         if (triggerbotJson.isMember("Key")) triggerbotConfig.key = triggerbotJson["Key"].asInt();
         if (triggerbotJson.isMember("Friendly fire")) triggerbotConfig.friendlyFire = triggerbotJson["Friendly fire"].asBool();
         if (triggerbotJson.isMember("Scoped only")) triggerbotConfig.scopedOnly = triggerbotJson["Scoped only"].asBool();
+		if (triggerbotJson.isMember("Hitchance enable")) triggerbotConfig.hitChance = triggerbotJson["Hitchance enable"].asBool();
+		if (triggerbotJson.isMember("Hitchance ammount")) triggerbotConfig.hitChanceAmmount = triggerbotJson["Hitchance ammount"].asInt();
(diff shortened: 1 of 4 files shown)


AbdallahMehiz:
Your code doesn't compile correctly, i think your code should be like this :
`            if (damage >= (config.triggerbot[weaponIndex].killshot ? trace.entity->health() : config.triggerbot[weaponIndex].minDamage)) { 
                if ((!config.triggerbot[weaponIndex].hitChance || (1.0f - (activeWeapon->getInaccuracy())) * 100.f >= config.triggerbot[weaponIndex].hitChanceAmmount)) {
                    cmd->buttons |= UserCmd::IN_ATTACK;
                    lastTime = 0.0f;
                }`
Instead of this:
`                if ((!config.triggerbot[weaponIndex].hitChance || (1.0f - (activeWeapon->getInaccuracy())) * 100.f >= hitchance)) {
                    cmd->buttons |= UserCmd::IN_ATTACK;
                    lastTime = 0.0f;
                }`
tell me if i am correct :")

pubsense:
I forgot to define the hitchance thing. added in last commit @AbdallahMehiz

pubsense:
> Your code doesn't compile correctly, i think your code should be like this :

forgot to add the float value for hitchance also to assign min value and max value for slider in gui

streamshark:
One thing I would note is max inaccuracy threshold and true hitchance are different things. Osiris doesn't currently have true hitchance which involves tracing a bunch of imaginary rays and seeing how many would hit to get a percent estimate of hitting the target, while max inaccuracy method just tells you the current inaccuracy of yourself based on if you are walking, crouched, etc. The major difference here is true hitchance will shoot if you are high inaccuracy but close to the target (likely to hit) for example, while max inaccuracy will shoot no matter the positioning of the target (more resource efficient but can lead to shooting when unlikely to hit). I suggest labeling this is as triggerbot max shot inaccuracy, in the same way it's labeled in the aimbot tab.

xAkiraMiura:
you could probably get away with calculating how large the player is in your view, if you were looking straight at them. a perspective calculation would probably be faster than ray tracing a lot.

AbdallahMehiz:
this might have to be updated because of the latestcommits from daniel

PR #1141, opened 2020-02-23
Friendly fire/team damage counter

xAkiraMiura:
When enabled in Misc, logs damage done to teammates in the developer console (`~`).
Example: `[ Friendly Fire] Kills: <2> Damage: <148>`

Keeps counting but doesn't log when the option is disabled (prevents desync)
Resets when the match starts, after warmup.

*also my vs plugins removed some trailing whitespace*

Built status with VC++ / VS2019
Debug x86: Success
Release x86: Success

Download with `git clone https://github.com/xAkiraMiura/Osiris.git -b team-damage-indicator`

### DIFF
--- a/Osiris/Hacks/Misc.cpp
+++ b/Osiris/Hacks/Misc.cpp
@@ -504,3 +504,27 @@ void Misc::killMessage(GameEvent& event) noexcept
     cmd += "\"";
     interfaces.engine->clientCmdUnrestricted(cmd.c_str());
 }
+
+void Misc::teamDamageCounter(GameEvent* event) noexcept {
+    if (!event || !interfaces.engine->isInGame()) return;
+
+    PlayerInfo pinfo;
+    interfaces.engine->getPlayerInfo(interfaces.engine->getLocalPlayer(), pinfo);
+
+    if (event->getInt("attacker") == pinfo.userId) { // we attacked.
+        if (!interfaces.entityList->getEntity(interfaces.engine->getPlayerForUserID(event->getInt("userid")))->isEnemy()) {
+            // teammate got hurt
+            switch (fnv::hashRuntime(event->getName())) {
+            case fnv::hash("player_hurt"):
+                teamDamage += event->getInt("dmg_health");
+                break;
+            case fnv::hash("player_death"):
+                teamKills += 1;
+                break;
+            }
+
+            if (config.misc.teamDamageCounter)
+                memory.debugMsg("[ Friendly Fire ] Kills: <%d> Damage: <%d>)\n", teamKills, teamDamage);
+        }
+    }
+}
--- a/Osiris/Hacks/Misc.h
+++ b/Osiris/Hacks/Misc.h
@@ -13,6 +13,9 @@
 
 namespace Misc
 {
+    inline int teamDamage = 0;
+    inline int teamKills = 0;
+
     void edgejump(UserCmd* cmd) noexcept;
     void slowwalk(UserCmd* cmd) noexcept;
     void inverseRagdollGravity() noexcept;
@@ -36,6 +39,7 @@ namespace Misc
     void fixTabletSignal() noexcept;
     void fakePrime() noexcept;
     void killMessage(GameEvent& event) noexcept;
+    void teamDamageCounter(GameEvent* event) noexcept;
 
     constexpr void fixMovement(UserCmd* cmd, float yaw) noexcept
     {
--- a/Osiris/Hooks.cpp
+++ b/Osiris/Hooks.cpp
@@ -228,7 +228,7 @@ static void __stdcall paintTraverse(unsigned int panel, bool forceRepaint, bool 
         Esp::render();
         Misc::drawBombTimer();
         Misc::spectatorList();
-        Misc::watermark();        
+        Misc::watermark();
         Visuals::hitMarker();
     }
     hooks.panel.callOriginal<void, 41>(panel, forceRepaint, allowForce);
@@ -327,13 +327,19 @@ static bool __stdcall fireEventClientSide(GameEvent* event) noexcept
     if (event) {
         switch (fnv::hashRuntime(event->getName())) {
         case fnv::hash("player_death"):
+            Misc::teamDamageCounter(event);
             Misc::killMessage(*event);
             SkinChanger::overrideHudIcon(*event);
             break;
         case fnv::hash("player_hurt"):
+            Misc::teamDamageCounter(event);
             Misc::playHitSound(*event);
-            Visuals::hitEffect(event);                
+            Visuals::hitEffect(event);
             Visuals::hitMarker(event);
+            break;
+        case fnv::hash("round_announce_match_start"):
+            Misc::teamKills = 0;
+            Misc::teamDamage = 0;
             break;
         }
     }
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -153,7 +153,7 @@ void Config::load(size_t id) noexcept
     for (size_t i = 0; i < esp.players.size(); i++) {
         const auto& espJson = json["Esp"]["Players"][i];
         auto& espConfig = esp.players[i];
-        
+
         if (espJson.isMember("Enabled")) espConfig.enabled = espJson["Enabled"].asBool();
         if (espJson.isMember("Font")) espConfig.font = espJson["Font"].asInt();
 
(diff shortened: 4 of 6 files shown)


streamshark:
looks cool, could be useful to track team damage / team kills other players on your team commit for griefing purposes as well in the future

xAkiraMiura:
I couldn't find a way to reset the variable whenever I join a new match so it only gets cleared at the start of one. This should be fine because you can't join mid-game in competitive but if you join before that you could possibly miss the start event.

barbadully:
would there be a way to make it colored so that you can see it easily in the console? also, do you have to bind a button to it to work or ?

xAkiraMiura:
whenever you shoot a teammate with it enabled it will log it to the console. i don't think player_hurt is networked when you shoot your teammates without friendly fire enabled. try going into a bot competitive match and shoot someone. i got it working there.

I know there is a colored varient of `memory.debugMsg` but i haven't found it.

Evilhelder:
NICE :D

Evilhelder:
Does it only show the damage you have done to your allies? Wouldn't it be possible to show others?

Ex: Player But  [ Friendly Fire ] Kills: <0> Damage: <8>
     Player Evil  [ Friendly Fire ] Kills: <-2> Damage: <240>

xAkiraMiura:
it only shows the damage that has been done in total to teammates, adding it for other players is possible but would require a lot of state management. you could probably make a vector where the index is the player index; you would have to shift it around whenever someone leaves or joins though.

player_hurt is always networked even if someone is dormant *i think*

Evilhelder:
hmmmm, I understand.

danielkrupinski:
> I know there is a colored varient of `memory.debugMsg` but i haven't found it.

Now you can use `Memory::conColorMsg` e.g.
```cpp
memory.conColorMsg({ 255, 255, 0, 255 }, "ColorTest");
```

Evilhelder:
> 
> 
> > I know there is a colored varient of `memory.debugMsg` but i haven't found it.
> 
> Now you can use `Memory::conColorMsg` e.g.
> 
> ```c++
> memory.conColorMsg({ 255, 255, 0, 255 }, "ColorTest");
> ```

Can you help with the code to show the damage caused by the team?

With any example or formula to do?

Evilhelder:
> 
> 
> I couldn't find a way to reset the variable whenever I join a new match so it only gets cleared at the start of one. This should be fine because you can't join mid-game in competitive but if you join before that you could possibly miss the start event.

wouldn't it be nice to create a "reset" button? So whenever you leave a match, you would reset the count.

xAkiraMiura:
added a reset counter button in the GUI.
i can't figure out how to rebase commits correctly so i don't have the concolormsg on my version.

Osiris/Hacks/Misc.cpp
```diff
@@ -504,3 +504,27 @@ void Misc::killMessage(GameEvent& event) noexcept
...
     interfaces.engine->clientCmdUnrestricted(cmd.c_str());
 }
+
+void Misc::teamDamageCounter(GameEvent* event) noexcept {
+    if (!event || !interfaces.engine->isInGame()) return;
+
+    PlayerInfo pinfo;
+    interfaces.engine->getPlayerInfo(interfaces.engine->getLocalPlayer(), pinfo);
+
+    if (event->getInt("attacker") == pinfo.userId) { // we attacked.
```
danielkrupinski: **interfaces.engine->getPlayerForUserID(event->getInt("attacker")) == interfaces.engine->getLocalPlayer()**
should work as well and does not involve PlayerInfo

Osiris/Hacks/Misc.cpp
```diff
@@ -504,3 +504,27 @@ void Misc::killMessage(GameEvent& event) noexcept
...
+        if (!interfaces.entityList->getEntity(interfaces.engine->getPlayerForUserID(event->getInt("userid")))->isEnemy()) {
```
danielkrupinski: Check for entity being null pointer, just in case

Osiris/Hacks/Misc.h
```diff
@@ -13,6 +13,9 @@
 
 namespace Misc
 {
+    inline int teamDamage = 0;
```
danielkrupinski: Should it really be in header? Maybe move into `Misc::teamDamageCounter` as a static variable.

Osiris/Hooks.cpp
```diff
@@ -327,14 +327,20 @@ static bool __stdcall fireEventClientSide(GameEvent* event) noexcept
...
             SkinChanger::overrideHudIcon(*event);
             break;
         case fnv::hash("player_hurt"):
+            Misc::teamDamageCounter(event);
             Misc::playHitSound(*event);
-            Visuals::hitEffect(event);                
+            Visuals::hitEffect(event);
             Visuals::hitMarker(event);
             break;
+        case fnv::hash("round_announce_match_start"):
```
danielkrupinski: This dispatch could be moved to `Misc::teamDamageCounter` in order to not involve `Misc::teamKills` and ` Misc::teamDamage` here.

Evilhelder:
YOU ARE THE BEST

Evilhelder:
@xAkiraMiura  The function is counting the damage done to itself ...

https://imgur.com/8dwxiHf

Osiris/Hooks.cpp
xAkiraMiura: currently `Misc::teamDamageCounter` relies on the event either being `player_hurt` or `player_death`.
either i could make a seperate function to set the variables or completely change the flow of `Misc::teamDamageCounter` to account for `round_announce_match_start`

Osiris/Hacks/Misc.h
xAkiraMiura: How would I convert it to a static variable? If I do then it is no longer changable in other functions. 
Would I make a function to reset it and call it instead of just setting them to zero?

xAkiraMiura:
added a check to prevent counting self damage @Evilhelder

Evilhelder:
> added a check to prevent counting self damage

I saw. Very good.

Evilhelder:
Now and find a way to count the damage from the rest of the team separately.

Showing nick and total damage ...

Ex:

image

effex1337:
Here's a tip look for server id/ip/name or whatever for unique identification of server so you can reset when u go to another game and not when you disconnect or have to click on a button.

xAkiraMiura:
> 
> 
> Here's a tip look for server id/ip/name or whatever for unique identification of server so you can reset when u go to another game and not when you disconnect or have to click on a button.

probably won't change whenever the round ends, i could watch for when isInGame() changes states but then I would be checking every tick rather than just being event driven, leading to slightly worse performance.

effex1337:
I think you can get the info on the players and stuff right after you connected.
If so then check for isConnected && !inGame like in namechanger.

effex1337:
Also moving this to a ui would be easier.

streamshark:
maybe a full imgui playerlist that uses team damage as a column would be the ideal implementation of this

Evilhelder:
> 
> 
> maybe a full imgui playerlist that uses team damage as a column would be the ideal implementation of this

If creating a window with the list of players, you could add "StealName", or "Fake Friendfire" ....

Evilhelder:
Or even a variation of the "reportbot" to flood only one player ...

Evilhelder:
FakeFriendly Fire Example: 
giphy

streamshark:
https://github.com/AimTuxOfficial/AimTux/blob/master/src/ATGUI/Windows/playerlist.cpp

aimtux did an implementation of a playerlist like this

effex1337:
Made in ImGui (alot cleaner and easier to use)
https://youtu.be/ltA4hGhS4bw

effex1337:
Oh i just read the codes. I thought it was something like this.
image

Evilhelder:
> Oh i just read the codes. I thought it was something like this.

xD. Osiris still working...

CS GO just updated and changed some things in the interface ...

effex1337:
https://youtu.be/z5mhgFVTS2o
An update on my grieving box
https://github.com/danielkrupinski/Osiris/pull/1149

Evilhelder:
> 
> 
> https://youtu.be/z5mhgFVTS2o
> An update on my grieving box
> #1149

NICE MAN, :D THIS IS BEAUTIFUL..

PR #1149, opened 2020-02-25
Added Friendly Fire/Kills/Grief Counter

noteffex:
Basically logs how many kills your team mates and you are on.
Helps grief/not get grieved.
In action:
https://youtu.be/z5mhgFVTS2o

### DIFF
--- a/Osiris/Config.h
+++ b/Osiris/Config.h
@@ -271,6 +271,7 @@ public:
         bool fixTabletSignal{ false };
         float maxAngleDelta{ 255.0f };
         bool fakePrime{ false };
+        bool griefBox{ false };
     } misc;
 
     struct {
--- a/Osiris/GUI.cpp
+++ b/Osiris/GUI.cpp
@@ -1064,6 +1064,15 @@ void GUI::renderMiscWindow(bool contentOnly) noexcept
     ImGui::SetNextItemWidth(120.0f);
     ImGui::SliderFloat("Max angle delta", &config.misc.maxAngleDelta, 0.0f, 255.0f, "%.2f");
     ImGui::Checkbox("Fake prime", &config.misc.fakePrime);
+    ImGui::Checkbox("Grief Box", &config.misc.griefBox);
+    if (config.misc.griefBox) {
+        if (ImGui::Button("Reset")) {
+            for (int i = 0; i < 65; i++) {
+                teamDamage[i] = 0;
+                teamKills[i] = 0;
+            }
+        }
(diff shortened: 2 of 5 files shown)


Evilhelder:
Question:

Do you move the window from place?

or makes it more transparent?

effex1337:
> Question:
> 
> Do you move the window from place?
> 
> or makes it more transparent?

Not sure what u meant but you can move it around.

streamshark:
looks good, I think evil was trying to ask if he can adjust window transparency of the box to see through it while playing

ClaudiuHKS:
Nice, but @danielkrupinski uses **MultiByteToWideChar** right after **GetPlayerInfo** for player names.

Evilhelder:
> 
> 
> looks good, I think evil was trying to ask if he can adjust window transparency of the box to see through it while playing

exactly

Evilhelder:
BUG: after switching sides it shows all players ..

image

xAkiraMiura:
When somebody leaves and another person takes their index then the new person would probably inherit the previous values.

Wasn't the point of this checking how close someone was to getting kicked? I think it would be a lot more useful to know that Bot Gabe is 70 damage away from getting kicked rather than knowning that I did 30 to Bot Gabe and 80 to Bot Jon.

steel4me:
Uhh, nice, are you interested to implement more? For example, showing matchmaking competetive wins of players. report / commend buttons behind colums and in future it could be the base for priority aimbot / trigger specific settings ;) 

Just some ideas.

AbdallahMehiz:
can this Gui be used with the spectator list?
image
would be a good thing to change for the spec list

xBaTm4n:
I have this problem: "Misc.cpp(30,40): error C2109: subscript requires array or pointer type"

danielkrupinski:
Misc::griefingBox() is not thread-safe and will cause crashes if engine simultaneously invalidates entity while we're using it. See https://github.com/danielkrupinski/GOESP/blob/70100d09a06ec33ead2531ea9f974ef0bc025a0a/GOESP/Hacks/Misc.cpp#L114 to understand how to make it stable.

CerberusGame:
https://prnt.sc/rymobq
I how to fix ?

Evilhelder:
FIX pls :( I BEG YOU. D:

braken1:
> can this Gui be used with the spectator list?
> image
> would be a good thing to change for the spec list

would be cool

PR #1176, opened 2020-03-01
Made Reportbot report players several times

AbdallahMehiz:


danielkrupinski:
I think it'd be better to loop through players n-times (report player 1, player 2 ...) than reporting one player multiple times in a row before moving to the next one.

AbdallahMehiz:
> I think it'd be better to loop through players n-times (report player 1, player 2 ...) than reporting one player multiple times in a row before moving to the next one.

It is better if there is a lot of players in a single server like in a community server(surf, bhop, jailbreak...etc) or in casual(deathmatch, dangerzone, wargames..etc), but i mostly use it on competitive and i think this is working fine.

ZerGo0:
I highly doubt that this is useful for anything, Valve themselves said in one of the last updates that people who abuse the report feature have a lower value report. Spamming reports will lower your report "score" and your enemies will most likely end up not getting banned because of that.

I can just confirm this, I have the reportbot enabled in my config and I have been playing quite a lot. I had spinners in the enemy team and just normal blatant cheaters. I haven't gotten any message for someone being banned in a while now probably a month or so now.

https://blog.counter-strike.net/index.php/2020/02/28450/
> Note that reports are weighted higher for players that play more and/or report less frequently, so accounts that generate no XP and/or spam reports will have little to no impact.

danielkrupinski:
I've implemented it differently, nonetheless thanks for your pull request.

PR #1262, opened 2020-03-15
Glow Material For Chams

effex1337:
Just a lot cleaner!


effex1337:
image
image

effex1337:
image
image

Chopard:
WTF?
FDY1aqErZ9E

Chopard:
WTF?
Screenshot_11

notgoodusename:
effex you should try this https://www.unknowncheats.me/forum/counterstrike-global-offensive/337040-dynamic-metallic-chams.html

danielkrupinski:
Great work! @Chopard you have to use glow as a secondary material.

effex1337:
Yes, it's supposed to be used as an overlay, meaning that you need to use a base material then apply glow as a second material.

PR #1317, opened 2020-03-25
Add max shot and aim inaccuracy to triggerbot

Vonr:
for now i dont see a reason for two settings there might be one but I'm too tired to look (there are 2 settings anyway)


xAkiraMiura:
one is for movement inaccuracy and the other is for spread inaccuracy.
some weapons like the negev only suffer from movement inaccuracy so keeping them together would help these weapons

Finndramere:
> one is for movement inaccuracy and the other is for spread inaccuracy.
> some weapons like the negev only suffer from movement inaccuracy so keeping them together would help these weapons

I think it doesn't work that way. Both have the same functionality one is limit to where aim will trigger and one is limit for when shot will be triggered(for autoshoot or while shooting). I think he's right that there's no point in having aim inaccuracy because well... You're not aiming when triggerbotting unless you implemented magnet trigger.  Correct me if I'm wrong.

from README.md
Max aim inaccuracy - maximum weapon inaccuracy allowing aimbot to run, lowering this value will e.g. disable aimbot while jumping or running.

Osiris/Hacks/Triggerbot.cpp
```diff
@@ -80,7 +80,7 @@ void Triggerbot::run(UserCmd* cmd) noexcept
         if (float armorRatio{ weaponData->armorRatio / 2.0f }; activeWeapon->itemDefinitionIndex2() != WeaponId::Taser && HitGroup::isArmored(trace.hitgroup, trace.entity->hasHelmet()))
             damage -= (trace.entity->armor() < damage * armorRatio / 2.0f ? trace.entity->armor() * 4.0f : damage) * (1.0f - armorRatio);
 
-        if (damage >= (config.triggerbot[weaponIndex].killshot ? trace.entity->health() : config.triggerbot[weaponIndex].minDamage) && activeWeapon->getInaccuracy() <= config.triggerbot[weaponIndex].maxShotInaccuracy) {
+        if (damage >= (config->triggerbot[weaponIndex].killshot ? trace.entity->health() : config->triggerbot[weaponIndex].minDamage) && activeWeapon->getInaccuracy() <= config.triggerbot[weaponIndex].maxShotInaccuracy) {
```
playday3008: config.triggerbot[weaponIndex].maxShotInaccuracy
you forgot "->" here

Vonr:
rewriting on newer branch

PR #1327, opened 2020-03-27
very small readme update

speedplane2248:
adds the glow material on chams


danielkrupinski:
Thank you for the pull request.

PR #1452, opened 2020-04-18
Add "Reload configs" button in Config tab

playday3008:
^


Oly-UK:
nice this is actually an awesome addition @danielkrupinski please pull this

notgoodusename:
what is this for?

w1ldac3:
For any of you not understanding, basically when you put someone's config or ur config in folder with settings, it won't refresh list of configs. So you need to re-inject cheat again. This fixes it.

danielkrupinski:
That's an incorrect approach, better do like in https://github.com/danielkrupinski/Osiris/blob/368806d35e666ca489d58bbe14a42210184f6218/Osiris/Config.cpp#L23-L26

PR #1482, opened 2020-04-24
Killsound

twangsriprasoet:


danielkrupinski:
Thank you for the contribution.

joshcancode:
Why not just combine both functions into one, and perform a check if it's being called from player_death or player_hurt?

twangsriprasoet:
For playing different sound between hit and kill. ;)

joshcancode:
> For playing different sound between hit and kill. ;)

I've already achieved this within one function.

`
void Misc::playDamageSound(GameEvent& event, bool playerHurt) noexcept
{
	auto& soundConfig = playerHurt ? config->misc.hitSound : config->misc.killSound;
	if (!soundConfig.enabled)
		return;

	PlayerInfo localInfo;
	if (!interfaces->engine->getPlayerInfo(localPlayer->index(), localInfo))
		return;

	if (event.getInt("attacker") != localInfo.userId || event.getInt("userid") == localInfo.userId)
		return;

	constexpr std::array sounds{
		"play physics/metal/metal_solid_impact_bullet2",
		"play buttons/arena_switch_press_02",
		"play training/timer_bell",
		"play physics/glass/glass_impact_bullet1"
	};

	if (static_cast<std::size_t>(soundConfig.sound) < sounds.size())
		interfaces->engine->clientCmdUnrestricted(sounds[soundConfig.sound]);
	else if (soundConfig.sound == 4)
		interfaces->engine->clientCmdUnrestricted(("play " + soundConfig.customSound).c_str());
}
`

Lets you play different sounds for hits and kills.

PR #1628, opened 2020-05-22
Make thirdperson show cmd->viewangles

DoomFishWasTaken:
Made this into a standalone pull because a lot of people have been requesting this to be a standalone pull.


notgoodusename:
Can be done easier and showing chocked viewangles wont make the player show your fake, just set it to the last sent viewangle rather than giving the option to the player, and setting the player viewangles to real will show your fake because the model isnt in sync with server animations

joshcancode:
A better way to do this would be to hook EyeAngles and modify the calls to pitch and yaw.

notgoodusename:
> A better way to do this would be to hook EyeAngles and modify the calls to pitch and yaw, it will also allow you to see your desync leaning.

i dont see how a modifying a netvar would make your angles visible.

joshcancode:
> > A better way to do this would be to hook EyeAngles and modify the calls to pitch and yaw, it will also allow you to see your desync leaning.
> 
> i dont see how a modifying a netvar would make your angles visible.

Don't hook the netvar, hook the function.

PR #1697, opened 2020-06-05
AimBOT's AutoWall Team Check Fix

ClaudiuHKS:
Blocks accidental team damage while AimBOT's friendly fire setting is turned off and visible only too.


danielkrupinski:
Thanks!

yannitraian:
am o problema si nu imi dau seama de ce.imi spune ....Filed to find VClient018.

PR #1712, opened 2020-06-08
Jump Check

zajkos:
I (with a lot of help from Kron1Q) add Jump Check to Aimbot

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -961,6 +961,7 @@ void Config::save(size_t id) const noexcept
         aimbotJson["Max shot inaccuracy"] = aimbotConfig.maxShotInaccuracy;
         aimbotJson["Min damage"] = aimbotConfig.minDamage;
         aimbotJson["Killshot"] = aimbotConfig.killshot;
+        aimbotJson["jump check"] = aimbotConfig.jumpCheck;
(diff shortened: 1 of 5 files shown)


notgoodusename:
no need to define onground just do flags() & 1

ClaudiuHKS:
Very good idea to automatically stop the aimbot while the ``localPlayer`` not on ground.

zajkos:
ok I will try to improve
PS. I'm just learning C ++ so for fun, I try to write something (but it doesn't always work out)

zajkos:
> no need to define onground just do flags() & 1

can u help ?  I'm really bad

dapaster:
**Aimbot.cpp**

if (!(localPlayer->flags() & 1) && config->aimbot[weaponIndex].jumpCheck) return;

**Entity.h**

Delete `#define FL_ONGROUND (1<<0)`

zajkos:
OK thanks, I have already corrected

zajkos:
BTW, it will be added or something else I have to do

techzin:
@danielkrupinski add this to the main branch.

dayz2000:
update ? pls ?

AntonDoil:
Works well, thanks.

PR #1731, opened 2020-06-11
Custom Skybox

caiocinel:
Just add skybox files in "csgo/materials/skybox", choose "Custom" in Skybox menu and write skyname in input.


xXTOO-HARDXx:
Good job.

ryseDMT:
Nice one, I added custom skyboxes to my build too, but my way was inferior to yours. I just added them to the source manually. Good job.

aljawary:
not work on official matchmaking : (

aljawary:
doesnt work in official matchmaking

barbadully:
I guess this wouldn't work in official mm, same as with the custom hitmarker sounds...

xXTOO-HARDXx:
@barbadully no, it's not the same case. All hitmarker sounds are custom, they aren't from csgo usually. Check the code

jestemgosciem:
dosen't work for me, help

cailloubr:
Ignore 11/Jun commits

jestemgosciem:
@danielkrupinski why custom skybox doesn't work on normal games(mm, wingman and others)?

cailloubr:
> @danielkrupinski why custom skybox doesn't work on normal games(mm, wingman and others)?

Perhaps https://developer.valvesoftware.com/wiki/Pure_Servers ?

xXTOO-HARDXx:
It won't be anything to do with Pure Servers, a 'pure server' is just a server that requires a TRUE value for sv_pure from each client that joins the server or their connection rejected. So as long as you're ON the server, a Pure Server isn't going to continuously check files or anything similar, afaik.

Cyk-Fad:
nice job

cailloubr:
Living and learning, ty.

PR #1750, opened 2020-06-13
Introducing Multipoints Technology

ClaudiuHKS:
I just introduced it, it doesn't change how AimBOT works.

Counter-strike  Global Offensive Screenshot 2020 06 13 - 13 12 38 82

### DIFF
--- a/Osiris/Hacks/Multipoints.cpp
+++ b/Osiris/Hacks/Multipoints.cpp
@@ -1,0 +1,170 @@
+#include "../Interfaces.h"
+#include "../Memory.h"
+
+#include "../SDK/Entity.h"
+#include "../SDK/Vector.h"
+#include "../SDK/matrix3x4.h"
+#include "../SDK/LocalPlayer.h"
+#include "../SDK/GlobalVars.h"
+#include "../SDK/ModelInfo.h"
+
+#include "../Hacks/Multipoints.h"
+
+namespace Multipoints
+{
+    bool retrieveAll(Entity* entity, float multiPointsExpansion, Vector(&multiPoints)[Multipoints::HITBOX_MAX][Multipoints::MULTIPOINTS_MAX])
+    {
+        if (localPlayer)
+        {
+            if (entity)
+            {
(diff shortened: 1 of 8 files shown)


dapaster:
Nice!

aljawary:
how it works ?

ClaudiuHKS:
> how it works ?

With SDK/ModelInfo.h
It's useful for the AimBOT feature. To have multiple hit points. Not just 8~ hit points. 171 (19x9).

techzin:
I think this should improve the accuracy of the aimbot?

notgoodusename:
yeah, but this doesnt change how the aimbot works in anyway

Osiris/Hacks/Multipoints.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "../SDK/GlobalVars.h"
+#include "../SDK/ModelInfo.h"
+
+#include "../Hacks/Multipoints.h"
+
+namespace Multipoints
+{
+    bool retrieveAll(Entity* entity, float multiPointsExpansion, Vector(&multiPoints)[Multipoints::HITBOX_MAX][Multipoints::MULTIPOINTS_MAX])
+    {
+        if (localPlayer)
```
notgoodusename, line 17: you can do if(!localPlayer)
return false 
instead of an if(localPlayer)
//code 
same aplies to the rest

techzin:
Is Need create another file to Multipoints? add this on ESP File.

zxsleebu:
i rly hope daniel will add this and will make it works (afaik you said it doesn't change aimbot)

zajkos:
Nice work !

Osiris/Hacks/Multipoints.cpp
ClaudiuHKS: You're right, I just wanted to have two 'return' only on each function. I will edit this soon.

xXTOO-HARDXx:
Very interesting system; not unique but still very cool, good job :) 👍

PR #1795, opened 2020-06-21
GUI Window Position SaveToConfig (lock/renember window position)

RyDeem:
As title says.

known bugs:
none at the moment

Its awesome to have window position just the way you like it without setup each gaming session.

Thoughts?

Found in GUI misc. tab bottom right.

The goal of this in my opinion would be a "lock" button/or icon next to the "X" that closes the window, to do what i have here but cleaner and on each window not in one place, but this works great nontheless currently.


notgoodusename:
you should make a new struct instead of using misc

RyDeem:
ok that makes sense, i did it.

RyDeem:
closed this because i messed up the pulls and this page shows random comments from dans pulls that should not be here etc.

Issue #1527, opened 2020-05-04
Glow Key


HackerGeo-sp1ne:
Can you add a glow with bind key?

Vadenez:
If you know anything bout C++, one of these might beable to help you:
 #1478
or maybe
#1294

danielkrupinski:
Added

PR #1844, opened 2020-07-03
 Add "On key" mode for glowing (related to #1527) 

Soniclev:
It supports multiple types of glow (allies, enemies and etc.) without any contradiction. 
demo


Cyk-Fad:
All Right

aljawary:
#1527

Issue #2007, opened 2020-08-02
(REQUEST) Add bloom remover


Frouk3:
I played in my server with sv_cheats 1 and with disabling bloom i get some fps

notgoodusename:
how do you remove bloom?

Frouk3:
> how do you remove bloom? (what convar)

mat_disable_bloom

RyDeem:
here you go @Frouk3 #2010

danielkrupinski:
Disabling postprocessing also disables bloom, doesn't it?

RyDeem:
I do think so dan heh. But i made it standalone if he wants it.

Frouk3:
mat_disable_bloom gives 5-10 fps but i don`t know how many fps gives in other pc

PR #2010, opened 2020-08-02
Added mat_disable_bloom...

RyDeem:
...toggle switch to visuals window below disable blur

Requested at #2007

Well i do belive that disable post-proccesing removes the bloom as well, but in anycase its standalone now, i do not notice a fps gain myself but maybe this helps someone.


RyDeem:
@notgoodusename I get the localplayer thing, i just added it for safety xd

oh i see now yes. i just kind of reused the ragdoll gravity code.

PR #2052, opened 2020-08-11
Allow using fake ban multiple times

enescglyn:


danielkrupinski:
It breaks name stealer.

enescglyn:
aw man

PR #2097, opened 2020-08-18
First shot delay and kill delay.

W4tev3n:


dayz2000:
what's frist shot delay mean ?
pre delay when aimbot start shot  ?

aljawary:
what do u mean can u xplain?

W4tev3n:
Yeah, first shot delay is a feature to let aimbot aim on a target without shooting, it's pretty useful for deagle, ssg and awp.
Kill delay waits a given while after killing a target before aiming on a next one. 
Here's the video demonstration of these features: [https://youtu.be/Blhx_BZOMIQ](url)

retart1337:
Get stickbugged lol

Rcsette:
Is that a way to make the aimbot works only after a seconds thats it is activated? I mean, I press the aimkey and it is activated secons after it, even if I am not in attack. thx

rad66 (approved):

PR #2112, opened 2020-08-22
Customizable Hitmarker

p-sw:
+ Length
+ Gap
+ Alpha
+ Color


sserve-kr:
> @danielkrupinski @sserve-kr Can you add a fade animation for the hitmarker?

@SapphireCodenz i'll try

sserve-kr:
made it, but not working.
idk why it isn't working, without any crash or compile error.
maybe because of my stupid brain

PR #2206, opened 2020-09-27
Cocksucker

cheaterbastard:


SwagsRGB:
Wow so edgy. You alone have entirely wiped the game of cheaters. However I would recommend learning Git first as just a pull request isn't committed without review and what you've done is just erased the read-me file. So even if your pull is merged, absolutely no harm to the cheat would be done lol.

danielkrupinski (commented):
Thanks for your pull request.

PR #2231, opened 2020-10-06
correct build and inject instructions

Ferryistaken:


newcommerdontblame:
Daniel would have added this if he wanted to because they made this. Current one is fine people can use whatever they want to inject

Ferryistaken:
you cannot use whatever injector. I tried xenos and project hacker 2 and they both didn't work

newcommerdontblame:
process hacker fixed after updates and never heard xenos before probably usong old method

Ferryistaken:
Maybe you should include that after you say to just use any injector? It's not like it can affect the project in a bad manner

bastquu:
@Ferryistaken mad?

Ferryistaken:
@bastquu  not really just trying to help the project. People who don't know shit about cheats won't be able to inject this

danielkrupinski (commented):
I do not want to advertise specific dll injector. Also launching as an administrator is not always mandatory.

Ferryistaken:
i just thought to put it there since I couldn't inject it and had to open an issue, some guy told me to use that injector and it finally worked

PR #1802, opened 2020-06-22
Bypass sv pure, Fake latency and better backtracking

notgoodusename:
For the new method of hooking

PR #2243, opened 2020-10-10
Custom skybox

Khajiitos:
_Edit: I just realised that this doesn't work in online servers, kinda sad tbh_

Add support of custom skyboxes.

Usage:
Drop your skybox files into csgo/materials/skybox/ directory (skybox consists of 12 files)
Select the "Custom" option from Skybox in Visuals tab
Type your custom skybox's name
And now you should have a custom skybox.

Example:

Skybox used in this example: sky_descent (https://gamebanana.com/textures/4895)

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -417,7 +417,7 @@ static void from_json(const json& j, Config::Visuals& v)
     read(j, "Flash reduction", v.flashReduction);
     read(j, "Brightness", v.brightness);
     read(j, "Skybox", v.skybox);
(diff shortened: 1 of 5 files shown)


newcommerdontblame:
u can do same thing with console if this only works in local game

newcommerdontblame:
Take a look sv_pure bypass (#1802) and try another

Khajiitos:
This also seems to be the case with custom hitsound. It doesn't work also spamming the console with messages
`[Sound] S_StartSound(): Failed to load sound '<soundname>'. File is missing from disk or is invalid.`

1nhale (approved):

PR #2259, opened 2020-10-18
Rainbow hud

Khajiitos:
Useless but cool

**Showcase:**
hud

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -427,6 +427,7 @@ static void from_json(const json& j, Config::Visuals& v)
     read(j, "Hit marker time", v.hitMarkerTime);
     read(j, "Playermodel T", v.playerModelT);
     read(j, "Playermodel CT", v.playerModelCT);
+    read(j, "Rainbow HUD", v.rainbowHud);
     read<value_t::object>(j, "Color correction", v.colorCorrection);
 }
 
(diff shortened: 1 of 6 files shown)


newcommerdontblame:
Add crosshair tho

Khajiitos:
> Add crosshair tho

I might do that tomorrow

SapphireCodenz:
ss?

bastquu:
cl_drawhud 0 and draw own hud, example https://imgur.com/Fz8x6g8

Khajiitos:
> cl_drawhud 0 and draw own hud, example https://imgur.com/Fz8x6g8

Using this command you would have to draw every single hud element by yourself, including crosshair and radar.

newcommerdontblame:
@Khajiitos bro it doesnt look smooth as crosshair pr also every color change makes radar flicking and it doesnt look that good 
Cant u use rgb instead of cvar currentcolor thing

Khajiitos:
> @Khajiitos bro it doesnt look smooth as crosshair pr also every color change makes radar flicking and it doesnt look that good
> Cant u use rgb instead of cvar currentcolor thing

No, because there are no rgb cvars for hud, only 10 presets, and I'm pretty sure that the radar flicking issue is caused by the game itself and I don't think we can do anything about it

PR #2272, opened 2020-10-25
Jump check for aimbot and triggerbot

newcommerdontblame:
1 line of code for not being too obvious with aimbot and not spamming in air with triggerbot


Khajiitos:
In my opinion it's better to just implement an inaccuracy check for triggerbot like aimbot has

newcommerdontblame:
> In my opinion it's better to just implement an inaccuracy check for triggerbot like aimbot has

there is triggerbot hitchance on my main too but its just for lplayer onground check that most of cheats has. Personally im using it for all weapons except ssg

newcommerdontblame:
also i think u will miss that shoot due to spread even if u on %90 hitchance. It's just pr that u can add if u want or no :))

dayz2000:
cant save load config

PR #2302, opened 2020-11-04
Change client ConVars

SquidmarcHD:
why is it fucking failing

### DIFF
--- a/Osiris/GUI.cpp
+++ b/Osiris/GUI.cpp
@@ -1203,7 +1203,6 @@ void GUI::renderStyleWindow(bool contentOnly) noexcept
         ImGui::End();
 }
 
-
 void GUI::renderMiscWindow(bool contentOnly) noexcept
 {
     if (!contentOnly) {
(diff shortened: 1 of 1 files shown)


Osiris/GUI.cpp
```diff
@@ -1365,7 +1366,16 @@ void GUI::renderMiscWindow(bool contentOnly) noexcept
         ImGui::EndPopup();
     }
     ImGui::PopID();
-
+    std::string cvar;
```
notgoodusename, line 1369: make a static string

SquidmarcHD: then itwould only change the first command entered

g1gantino: nice men1! i voted for you !!! hope yo u win election!!

Cyk-Fad: and when it has set, reset this string value

techzin:
the convart set value need a int or bool value, not string.

PR #2347, opened 2020-11-25
Auto disconnect

Khajiitos:
Automatically disconnects from the game when the match ends


0xb77a9:
Sound useless sorry
but nice work!

daisuki333:
Are you able to recreate another version of this but instead if someone tries to vote kick you it disconnects which is basically like an anti-vote kick thing. I'm not sure exactly how that works but that would be cool.

Khajiitos:
> Are you able to recreate another version of this but instead if someone tries to vote kick you it disconnects which is basically like an anti-vote kick thing. I'm not sure exactly how that works but that would be cool.

This is pointless as even if you disconnect, if the vote is successful you won't be able to join back

thisisprettyepic:
Nice! Thanks!

wemanzoz:
How do i add this to my osiris build

Heyter:
Useful, thanks.

PR #2357, opened 2020-12-01
Merge linux branch

danielkrupinski:


techzin:
Hmm, is now working with Windows and Linux right?

PR #2359, opened 2020-12-01
Viewmodel roll

Khajiitos:


thisisprettyepic:
you’re going in with these updates, keep it up

debooga:
don't use DWORD because its defined in windows.h and windows only

debooga:
u dont need to cast it in the first place, btw the rest of the code is great

PR #2418, opened 2020-12-19
Fix struct problem in SkinChanger::getQualities()

muhammetyurdan:
It was giving an error in compilation. I haven't tested the code in game.


caiocinel:
Fixed with 8cb516a

PR #2455, opened 2020-12-28
Adjustable hit sound volume

grey-stack:
Special thanks to @blackMasterLord for the help.


doggobot-owner:
Error, cant convert const char [24] into const size_t

grey-stack:
> Error, cant convert const char [24] into const size_t

https://github.com/grey-stack/Osiris-VolumeSlider/blob/a17460daf160103d06812f084d4de26df21b2713/Osiris/Hacks/Misc.cpp#L752-L759

try replacing the "snprintf" functions with the commented "sprintf_s" functions, hope it helps!

PR #2465, opened 2021-01-01
Implement Boyer-Moore algorithm for memory pattern searching

danielkrupinski:
Improves Memory class initialization performance by ~35%


degeneratehyperbola:
I don't know, it actually worsened injection times a lot.

danielkrupinski:
My measurements showed reduction of execution time from 0.22 sec to 0.16 sec.

PR #2518, opened 2021-01-22
Implement traceToExit for linux

SM9CC:
Hi, can you take a look at this, I attempted to implement this based on how Fuzion does it because I noticed the ASM would only work with Windows and auto wall is broken on linux currently.


SM9CC:
This works btw.

mkot2 (approved):

SM9CC:
https://github.com/danielkrupinski/Osiris/pull/3088

PR #2524, opened 2021-01-25
Molotov hull

danielkrupinski:


doggobot-owner:
bad

linuxer9:
> 
> 
> bad

why ? 
it's a helpful feature to see the exact extent of fire

PR #2558, opened 2021-02-06
Add more UTF-8 to ImGui Glyph Ranges

ClaudiuHKS:
To support Sunset Storm 壱 and Sunset Storm 弐 paint kits.

### DIFF
--- a/Osiris/Helpers.cpp
+++ b/Osiris/Helpers.cpp
@@ -53,7 +53,7 @@ ImWchar* Helpers::getFontGlyphRanges() noexcept
         builder.AddRanges(baseRanges);
         builder.AddRanges(ImGui::GetIO().Fonts->GetGlyphRangesCyrillic());
         builder.AddRanges(ImGui::GetIO().Fonts->GetGlyphRangesChineseSimplifiedCommon());
-        builder.AddText("\u9F8D\u738B\u2122");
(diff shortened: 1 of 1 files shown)


wemanzoz:
kinda swag

mkot2:
ok

mkot2 (approved):

danielkrupinski (commented):
Tahoma font which is used in GUI doesn't contain those characters, so they still won't appear correctly. There is also a `M4A4 | 龍王 (Dragon King)` skin which has chinese characters in its name.

ClaudiuHKS:
> Tahoma font which is used in GUI doesn't contain those characters, so they still won't appear correctly. There is also a `M4A4 | 龍王 (Dragon King)` skin which has chinese characters in its name.

You're right, I've noticed that. I've just created this **PR** as a _completion_. Of course, I don't expect `100 %` to be merged. Can be closed when the entire issue gets fixed. Maybe you should replace the font with a more **complex** one. 

`/Steam/steamapps/common/Counter-Strike Global Offensive/platform/vgui/fonts/xarialuni.ttf`, for example. Your decision, anyway.

danielkrupinski (commented):
Now **Desert Eagle | Sunset Storm 弐**, **Desert Eagle | Sunset Storm 壱** and **M4A4 | 龍王 (Dragon King)** display correctly.

PR #2597, opened 2021-02-20
Reimplement "Watermark" using imgui

danielkrupinski:


shakeit1337:
Hey @danielkrupinski how to add tick & cpu to this watermark ?

PR #2650, opened 2021-03-02
Update README.md

ElJaviLuki:
Help solve a very common issue for Osiris cheat novice users.


mkot2:
how about telling them to use a manual map injector

PR #2723, opened 2021-03-22
Add vote revealer

mkot2:
didn't test this so tell me if there are any problems

it outputs to console right now and it's just fine, but is there a way to output messages to chat and only have local player see them?

### DIFF
--- a/Osiris/Hacks/Misc.cpp
+++ b/Osiris/Hacks/Misc.cpp
@@ -950,7 +950,8 @@ void Misc::preserveKillfeed(bool roundStart) noexcept
     }
 }
 
-void Misc::voteRevealer(GameEvent& event) {
+void Misc::voteRevealer(GameEvent& event) noexcept
+{
     if (!config->misc.revealVotes)
         return;
 
--- a/Osiris/Hacks/Misc.cpp
+++ b/Osiris/Hacks/Misc.cpp
@@ -954,8 +954,8 @@ void Misc::voteRevealer(GameEvent& event) {
     if (!config->misc.revealVotes)
         return;
 
-    auto entity = interfaces->entityList->getEntity(event.getInt("entityid"));
-    if (!entity)
+    const auto entity = interfaces->entityList->getEntity(event.getInt("entityid"));
+    if (!entity || !entity->isPlayer())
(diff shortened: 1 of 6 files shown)


Tothale:
I have tried in gun race, working good.

cristyanul:
It works!

KernCS:
> but is there a way to output messages to chat and only have local player see them?

https://www.unknowncheats.me/forum/counterstrike-global-offensive/171375-printing-text-chat.html

Osiris/Hacks/Misc.cpp
```diff
@@ -950,6 +950,28 @@ void Misc::preserveKillfeed(bool roundStart) noexcept
...
 
+void Misc::voteRevealer(GameEvent* event = nullptr) {
+    if (!config->misc.revealVotes)
+        return;
+
+    if (event) {
+        switch (fnv::hashRuntime(event->getName())) {
+        case fnv::hash("vote_cast"):
+            //dont ask me why this is here
+            std::ostringstream str;
```
danielkrupinski, line 961: No need to use `std::ostringstream`.

Osiris/Hacks/Misc.cpp
```diff
@@ -950,6 +950,28 @@ void Misc::preserveKillfeed(bool roundStart) noexcept
...
+            str << interfaces->entityList->getEntity(event->getInt("entityid"))->getPlayerName() << " : ";
```
danielkrupinski, line 962: Entity nullptr check is missing.

Osiris/Hacks/Misc.cpp
```diff
@@ -950,6 +950,28 @@ void Misc::preserveKillfeed(bool roundStart) noexcept
     }
 }
 
+void Misc::voteRevealer(GameEvent* event = nullptr) {
```
danielkrupinski, line 953: `event` could be `GameEvent&` as the function does nothing when `event` is nullptr.

danielkrupinski (changes requested):

danielkrupinski (approved):

danielkrupinski:
Thank you for the contribution!

PR #2787, opened 2021-04-12
Impl SendPacket for Linux

SM9CC:
This fixes choked packets for Windows & Linux.


SM9CC:
Can somebody help a noob? Not sure why its not building on Windows and I don't have a Windows VM running currently.

Osiris/SDK/NetworkChannel.h
```diff
@@ -4,8 +4,9 @@
 
 class NetworkChannel {
 public:
-    VIRTUAL_METHOD(float, getLatency, 9, (int flow), (this, flow))
+    VIRTUAL_METHOD(float, getLatency, IS_WIN32() ? 9 : 8, (int flow), (this, flow))
+
+    std::byte pad[WIN32_LINUX(43, 63)];
```
danielkrupinski, line 9: `NetworkChannel::chokedPackets` is 4-byte aligned, so this isn't correct.

SM9CC: Sorry for my mistakes but when I tested it on Linux it at least seems to be working, Increasing / decreasing the choked packets did feel correct (More lag as I increased, and less when decreased)

My understanding was that `std::byte pad` had a size of 1 + (43 / 63) which makes `NetworkChannel::chokedPackets` 44 (0x002C) on Windows and 64 (0x0040) on Linux?

Linux Reference - https://github.com/LWSS/Fuzion/blob/master/src/SDK/INetChannel.h#L177

Windows Reference - https://github.com/AllusiveWheat/skeet.pw/blob/cb77743dad6abec5acc1fbbdc1a1f7d38c0165a3/sdk/interfaces/IVEngineClient.hpp#L351

Very new to this stuff, would appreciate if you could explain a little.

Thanks.

noche-x: I think this could use `__builtin_frame_address()` for Linux and `_AddressOfReturnAddress()` for Windows right?

In my opinion it simplifies things.

danielkrupinski: It works due to struct alignment which aligns `NetworkChannel::chokedPackets` to 4-byte offset but it's misleading to have 43/63 bytes padding. When you want your field to be at **std::uintptr_t(this) + 0x40** that means you need 64-byte padding.

Osiris/SDK/NetworkChannel.h
```diff
@@ -7,8 +7,8 @@ class NetworkChannel {
 public:
     INCONSTRUCTIBLE(NetworkChannel)
 
-    VIRTUAL_METHOD(float, getLatency, 9, (int flow), (this, flow))
+    VIRTUAL_METHOD(float, getLatency, IS_WIN32() ? 9 : 8, (int flow), (this, flow))
```
danielkrupinski, line 10: Index of 9 is correct on linux too.

danielkrupinski (changes requested):
Please correct padding size and `NetworkChannel::getLatency()` index.

PR #2788, opened 2021-04-12
Cathook gdb, Toolbox script, Wait for serverbrowser

SM9CC:
Added cathook gdb, Toolbox script (Thanks gamesneeze), Wait for serverbrowser to allow early injection.


wemanzoz:
what does this do exactly

SM9CC:
You can use the script to load, build etc on Linux, the cathook gdb which is included works around a bug on Arch Linux, the wait for server browser allows the script to inject as soon as the csgo process becomes available without crashing.

Osiris/Hooks.cpp
```diff
@@ -712,6 +713,12 @@ static int pollEvent(SDL_Event* event) noexcept
 
 Hooks::Hooks() noexcept
 {
+#ifdef _WIN32
+    while (!GetModuleHandleA("serverbrowser.dll")) std::this_thread::sleep_for(std::chrono::milliseconds(50));
```
mkot2, line 717: this Hooks::Hooks is only run on linux, put this into the other Hooks::Hooks

SM9CC: Done thanks!

cstles:
u sure this would work? it uses a dll

PR #2806, opened 2021-04-19
Remove includes suggested by CLion - 'Detect not directly used'

SM9CC:
Detect not directly used (default): this strategy follows the Include What You Use principle (the principle that if you use a symbol or type from a header, you should include that header) and detects the #includes directives with the declarations not used in the current file directly.

https://www.jetbrains.com/help/clion/unused-include-directive.html

### DIFF
--- a/Osiris/Config.cpp
+++ b/Osiris/Config.cpp
@@ -2,7 +2,6 @@
 #include <cstdlib>
 #include <fstream>
 #include <iomanip>
-#include <iterator>
 #include <system_error>
 
 #ifdef _WIN32
--- a/Osiris/GUI.cpp
+++ b/Osiris/GUI.cpp
@@ -1,11 +1,8 @@
-﻿#include <algorithm>
-#include <array>
+﻿#include <array>
 #include <cwchar>
 #include <fstream>
-#include <iterator>
 #include <string>
 #include <string_view>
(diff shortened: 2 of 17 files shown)


Osiris/Config.cpp
```diff
@@ -2,7 +2,6 @@
 #include <cstdlib>
 #include <fstream>
 #include <iomanip>
-#include <iterator>
```
danielkrupinski, line 5: `std::back_inserter()` is defined in `<iterator>`

Osiris/GUI.cpp
```diff
@@ -1,11 +1,8 @@
-﻿#include <algorithm>
-#include <array>
+﻿#include <array>
 #include <cwchar>
 #include <fstream>
-#include <iterator>
```
danielkrupinski, line 5: `std::distance()` is defined in `<iterator>`

Osiris/GUI.cpp
```diff
@@ -1,11 +1,8 @@
-﻿#include <algorithm>
```
danielkrupinski, line 1: `std::clamp()` is defined in `<algorithm>`

danielkrupinski:
Looks like CLion removes headers that are actually used in source files.

PR #2829, opened 2021-04-25
Allow Fake Prime on Linux

mdilai:
See no reason for it to be disabled on Linux

### DIFF
--- a/Osiris/Hacks/Misc.cpp
+++ b/Osiris/Hacks/Misc.cpp
@@ -6,6 +6,11 @@
 #include <sstream>
 #include <vector>
 
+#ifdef __linux__
+#include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


cstles:
thing is that its likely just disabled because of the fact that it might need to use some windows functions or dynamic link libraries

danielkrupinski (approved):

mdilai:
> thing is that its likely just disabled because of the fact that it might need to use some windows functions or dynamic link libraries

Em... no, it just one byte in memory, same as on Windows, there is just slightly different way to override memory on Windows and POSIX systems. Think this can be realized as universal macros but i see no reason to bother with it.

danielkrupinski:
Thanks.

PR #2831, opened 2021-04-26
Cross-platform calculateVmtLength() function

mdilai:
1. Moved calculateVmtLength() function to Helpers.cpp for code reuse
2. Implemented proper calculation for Linux
3. Changed MinHook, VmtHook and VmtSwap to call this function instead of duplicated code in three places.

### DIFF
--- a/Osiris/Helpers.cpp
+++ b/Osiris/Helpers.cpp
@@ -7,6 +7,13 @@
 #include <fstream>
 #include <string_view>
 
+#ifdef _WIN32
+#include <Windows.h>
+#else
+#include <dlfcn.h>
+#include <sys/mman.h>
+#endif
+
 #include "imgui/imgui.h"
 
 #include "ConfigStructs.h"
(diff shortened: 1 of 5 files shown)


cristeigabriel:
Can't we calculate a virtual method table's size just like this?
```
    while (m_original_table[m_table_size] != 0) {
        ++m_table_size;
    }
```

mkot2 (approved):

mkot2:
> 
> 
> Can't we calculate a virtual method table's size just like this?
> 
> ```
>     while (m_original_table[m_table_size] != 0) {
>         ++m_table_size;
>     }
> ```

if it was that simple why would he do it like this with OS specific things

cristeigabriel:
> > Can't we calculate a virtual method table's size just like this?
> > ```
> >     while (m_original_table[m_table_size] != 0) {
> >         ++m_table_size;
> >     }
> > ```
> 
> if it was that simple why would he do it like this with OS specific things

Yeah, that's what I'm asking, the method I've presented works on my end.

mdilai:
> > > Can't we calculate a virtual method table's size just like this?
> 
> > > ```
> 
> > >     while (m_original_table[m_table_size] != 0) {
> 
> > >         ++m_table_size;
> 
> > >     }
> 
> > > ```
> 
> > 
> 
> > if it was that simple why would he do it like this with OS specific things
> 
> 
> 
> Yeah, that's what I'm asking, the method I've presented works on my end.

To be honest, i am not sure that this is the easiest way to do that and your code _looks_ working too. What i did is ported exactly behavior of Windows code to Linux. Probably this should be more safer way with negligible performance effects, but i am not sure, let Daniel decide the future of this PR. I just was unable to watch this hardcoded variable anymore.

danielkrupinski (commented):
The problem is that `mprotect()` creates an additional rwx section:
#### Before injecting
before
#### After
after

mdilai:
@danielkrupinski understand. So lets stick to @cristeigabriel method, it work for me without any issues.

danielkrupinski (approved):

danielkrupinski:
Thanks!

PR #2841, opened 2021-05-01
[LINUX] Building process fixes and optimization, huge binary size reduction.

mdilai:
Maybe a better to make a few smaller pull requests instead of one large, but i feel this as one work, there is too many opened pull requests, i'm afraid of merge conflicts if PRs will be merged one by a time and just too lazy. Here is explanation of each of 5 commits:

1. **Using distro-provided freetype2 library on Linux instead of in-tree version**
Looks like in-tree version of Freetype2 hasn't any customizations and just added as dependency of ImGUI Truetype library. Although this is a good solution for Windows, any Linux distro already have freetype2 installed and it is a dependency of CSGO itself so there is no reason to link static Freetype2 inside binary, which not only waste a third of whole compiled binary size, but can also lead to rendering inconsistency between static freetype on Osiris GUI and native/steam-runtime libfreetype.so.6 used by rest of CSGO UI.

2. **Add freetype-dev dependency for github workflow on Linux**
Using of dynamic-linked freetype add one additional build dependency, but it whole size is smaller than overhead of static freetype added to every Osiris build, so i think this is a good deal.

3. **Using SDL provided OpenGL loader instead of custom one**
As we injecting in SDL application and using ImGUI SDL platform binder, SDL had already initialized OpenGL and exported all needed glExt prototypes so there is no need to bloat Osiris with GL3W and waste disk, memory and cpu time for it. Its crazy how ImGUI hasn't this option as one of many predefined loaders, although it need only three lines of code to use.
As a bonus, as we no more calling glExt functions directly, linker is able to link cheat against modern lightweight libOpenGL.so instead of legacy libGL.so & libGLX.so. Of course, if we provide linker with all available options to decide (see item №5.2)

4. **Do not compile imgui_demo.cpp**
I have read comments on beginning of file and agree on leaving this file in-tree as live documentation, but i see no reason to spent time first to compile this demo file and then to eliminate it by LTO on every build. I just removed it from both CMakeLists.txt and Osiris.vcxproj(.filters) but didn't touched the file itself.

5. **Fixing CMakeLists.txt to use proper CMake autodetection instead of hardcoded values**
It's a pain to look at current CMakeLists.txt which collected almost all "do not do" from CMake docs and ignoring all powerful of CMake, treating it like a primitive GNU Makefile (which it is not). Issues i've decided to fix following issues:
- **Not using CMake to check presence of build dependencies.** Now CMake will check if system have SDL2, Freetype and OpenGL libs/headers installed before build and show nice missing package error instead of typical hundreds of meaningless C++ compiler errors due one missing header
- **Using hardcoded include dirs and link libraries**, which not portable, specified based on Ubuntu 20.04 and can become broken at anytime "tomorrow" for rolling-based distros. Now we let CMake to find all needed include dirs and choose best libraries to link (for example libOpenGL.so instead of libGL.so/libGLX.so on modern distros with GLVND, see item №2)
- **Documentation discourage usage of include_directories() or link_directories() without a target specifier**. And we don't need link_directories() at all as CMake is smart enough to look at /usr/local/lib if needed.
- **Documentation discourage usage of target_link_libraries() without a visibility specifier**. Have set them as PRIVATE.

Here is result of optimization:
Binary size before: 1598256 bytes
Binary size after: 1041136 bytes

It's a 35% reduction of binary size + faster compiling time + slightly faster injection and initialization of cheat (on 10-years old AMD Phenom II, probably unnoticeable on modern CPUs). Hope i did all right and PR will be merged.


danielkrupinski (changes requested):
`imgui_demo.cpp` is really useful during development so undo aea8118d049ad3de0edfc001ce62b4655d83cc48 please.

danielkrupinski (changes requested):
When using OpenGL package Osiris links with libOpenGL.so, which isn't present in process of the game and needs to be loaded. It's better to use libGL.so which is used by the game itself.

CMakeLists.txt
```diff
@@ -15,35 +15,16 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
-link_directories(/usr/local/lib)
+find_package(Freetype REQUIRED)
+find_package(SDL2 REQUIRED)
+find_package(OpenGL REQUIRED)
 
 add_library(Osiris SHARED ${SOURCES})
 
-target_link_libraries(Osiris SDL2 GL dl)
+target_include_directories(Osiris PRIVATE ${SDL2_INCLUDE_DIRS} ${FREETYPE_INCLUDE_DIRS})
+target_link_libraries(Osiris PRIVATE SDL2 OpenGL::OpenGL ${CMAKE_DL_LIBS} Freetype::Freetype)
```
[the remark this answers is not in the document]
mdilai: Imported namespaces is a preferred way to link libraries since CMake 3.10. Macros even started removing from latest versions what unfortunately break backward compatibility for SDL2. So
`${SDL2_LIBRARIES}` 
Works only in Ubuntu, which, for some reason, removed original cmake module in their "libsdl2-dev" and replaced it with backported old "sdl2-config.cmake" which export only this macros. However, Debian and Arch using new SDL2Config.cmake/SDL2Targets.cmake modules, which export SDL2 namespace with `SDL2::SDL2` and `SDL2::SDL2main` targets, but not providing `${SDL2_LIBRARIES}`  at all. So adding just "SDL2" to target_link_libraries is not the most beautiful but the most portable way until Ubuntu 22.04 will be released.

danielkrupinski (approved):

CMakeLists.txt
```diff
@@ -15,35 +15,18 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
...
+
+find_package(Freetype REQUIRED)
+find_package(SDL2 REQUIRED)
+find_package(OpenGL REQUIRED)
 
 add_library(Osiris SHARED ${SOURCES})
 
-target_link_libraries(Osiris SDL2 GL dl)
+target_include_directories(Osiris PRIVATE ${SDL2_INCLUDE_DIRS} ${FREETYPE_INCLUDE_DIRS})
+target_link_libraries(Osiris PRIVATE SDL2 OpenGL::GL ${CMAKE_DL_LIBS} Freetype::Freetype)
```
danielkrupinski, line 31: What about just linking with `GL` here as it was before without using OpenGL package?

mdilai: I can ask why you recently replaced all stl::find with stl::ranges::find so Osiris not more compiling with LLVM's libc++(which not yet support STL ranges). Let me guess that it's because Osiris is a self-education project for you and you trying to use most modern programming practices in your code. With this feelings i've coded CMakeLists.txt using modern CMake syntax. But sure, it's your project and if you feel that it should link against "-lGL" directly i can make this, but, in my opinion, my way is more correct and future-proof.

danielkrupinski: > I can ask why you recently replaced all stl::find with stl::ranges::find so Osiris not more compiling with LLVM's libc++(which not yet support STL ranges).
We have a GitHub Action that uses clang-10 and compiles successfully.

Using `find_package()` and setting a variable is overcomplicating the process in my opinion.

mdilai: I was talking about [LLVM libc++](https://libcxx.llvm.org/), AFAIK it used by default only on OSX (where must be much more problems to run Osiris) and it should be especially enabled on Linux. I just test tested it and it broken due stl::ranges still not supported. Not a big deal anyway.

mdilai:
Ok, i have did some extreme testing with custom and cross-compilers and was unable to get any issues - CMake is really smart enough to provide `-L/usr/lib64` even if compiler instructed to not try to search libraries there. So i think we can revert to linking with just `GL` as a compromise, because Freetype and GL3W changes are much more important than CMakeLists.txt aesthetic.

By the way, during testing i found a small issue with CMAKE_CXX_STANDARD. According to [docs](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html)

> If the value requested does not result in a compile flag being added for the compiler in use, a previous standard flag will be added instead. This means that using:
> 
> set_property(TARGET tgt PROPERTY CXX_STANDARD 11)
> with a compiler which does not support -std=gnu++11 or an equivalent flag will not result in an error or warning, but will instead add the -std=gnu++98 flag if supported.

This mean that if we try to compile Osiris with compiler that don't support C++-20 (most LTS distros like RHEL Desktop and Debian Stable bundled with GCC8.3) we will get a lot of compiler errors because unsupported syntax. This can be easily fixed using `set(CMAKE_CXX_STANDARD_REQUIRED ON)` so we get a nice notification about compiler incompatibility.

I hope this is a final commit before merging.

danielkrupinski:
Thank you!

danielkrupinski (approved):

MrOpxDRS:
@mdilai Can You give me .so file for linux i have the load :) i dont know how to compile it many errors

PR #2844, opened 2021-05-01
Fix possible dx9 memory leaks

ClaudiuHKS:
After `Capture()`, `Release()` must be called. See [Remarks](https://docs.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dstateblock9-capture#remarks).

[Once again, since the stateblock object is an interface, you will need to release it when you are done with it.](https://docs.microsoft.com/en-us/windows/win32/direct3d9/state-blocks-save-and-restore-state#capture-individual-states)

See also, [the original imgui_impl_dx9.cpp#L150](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_dx9.cpp#L150).


mkot2 (approved):

danielkrupinski:
Thanks!

PR #2853, opened 2021-05-03
Added build and inject scripts and instructions for ubuntu and arch linux

hugohabicht01:


load_osiris
```diff
@@ -0,0 +1,75 @@
...
+  -ex "set logging on" \
+  -ex "set logging file /dev/null" \
+  -ex "set logging redirect on" \
+  -ex "set auto-load safe-path /usr/share/gdb/auto-load/usr/$lib_dir_name/:/usr/$lib_dir_name/" \
+  -ex "attach $csgo_pid" \
+  -ex "set \$dlopen = (void*(*)(char*, int)) dlopen" \
+  -ex "call \$dlopen(\"/usr/lib/$filename\", 1)" \
+  -ex "detach" \
+  -ex "quit"
+)"
```
mdilai, line 61: Oh, copy-paste from LWSS, how nice but unfortunately incorrect.

1. Logging options in wrong order so we will still creating and enlarging gdb.txt inside running folder. 
2. Auto-load safe-path point to /usr/lib64 what throwing errors on systems using /usr/lib instead of /usr/lib64. And we just don't need to auto-load anything at all.
3. dlopen() prototype incorrect and function unnecessarily complicated.

All this should be replaced with something like
```
input="$(
sudo gdb -n -q -batch-silent \
       -ex "set logging file /dev/null" \
       -ex "set logging redirect on" \
       -ex "set logging on" \
       -ex "attach $csgo_pid" \
       -ex "call ((void*(*)(const char*, int))dlopen)(\"/usr/lib/$filename\", 1)"   \
       -ex "detach" \
       -ex "quit"
)"
```

Elkellympia:
what happened to this? why is this still open? without this there is not really a way of running osiris on linux, does these scripts still work?

PR #2858, opened 2021-05-04
dx9 fixed build error

Heyter:
imgui_impl_dx9.cpp(301,54): error C2039: "uint32_t": is not a member of "std"


mkot2:
never got this error

danielkrupinski:
Thank you.

PR #2956, opened 2021-06-04
Remove fake prime

mkot2:
after last update this does nothing
you probably would have done this anyways but oh well


voidzero-development:
Is it patched for sure?

mkot2:
> 
> 
> Is it patched for sure?

with the last update, even if there was a way to fake prime (which there probably wont be) it would be useless

PR #2968, opened 2021-06-08
Close menu on Escape key

qualterz:
A more convenient way to close the menu


wemanzoz:
why

qualterz:
> 
> 
> why

I think it will improve usability

wemanzoz:
> > 
> > 
> > why
> 
> I think it will improve usability

i mean u can bind it to del or pgdn or pgup

mkot2:
you can change the keybind in config, also almost every cheat has the menu key on insert and nobody has a problem with it

qualterz:
Closing a popup menu with the Escape key is more intuitive than using a different key on the other side of the keyboard, especially when setting up a config for the first time. It is just good UX practice

mkot2:
> 
> 
> Closing a popup menu with the Escape key is more intuitive than using a different key on the other side of the keyboard, especially when setting up a config for the first time. It is just good UX practice

the first time user should read the getting started section in which it clearly states "menu key is INSERT", why should there be two different keys, also again almost every cheat does this too

OTGCable:
surely you're aware that the escape key is also the cs:go pause button, correct?
the entire idea behind using the escape/del/insert/home keys is to use keys that are not in use by the game.

qualterz:
> surely you're aware that the escape key is also the cs:go pause button, correct?
> the entire idea behind using the escape/del/insert/home keys is to use keys that are not in use by the game.

Yes, but in the context of a menu, the Escape key should mean closing it
I may not understand the philosophy of cheats, but from UX point of view, it is more convenient

qualterz:
> > Closing a popup menu with the Escape key is more intuitive than using a different key on the other side of the keyboard, especially when setting up a config for the first time. It is just good UX practice
> 
> the first time user should read the getting started section in which it clearly states "menu key is INSERT", why should there be two different keys, also again almost every cheat does this too

The Insert key to toggle the menu, Escape key to close it. The Escape key is optional, but for some users it will be useful

R3nzTheCodeGOD:
Might be the most useless feature I've ever seen

Asviix:
Useless or not, let the man train his coding and share what HE thinks is useful to HIM
stop hating and just be thankful he shares his work.

techzin (approved):
Yeah, can put on source.

techzin:
> Might be the most useless feature I've ever seen

I believe that the reason for an IT, to use windows so fast, is because it knows how to take useless shortcuts that others do not know.

CTRL + SHIFT + ESC

techzin:
This makes it simpler, I think the distance from the esc button is smaller than Insert.

techzin:
@qualterz  I imagine a little problem in theory, if ESC is now included in keycode list, then will the hotkey also understand ESC as a key, or as a cancel?

qualterz:
> 
> 
> @qualterz I imagine a little problem in theory, if ESC is now included in keycode list, then will the hotkey also understand ESC as a key, or as a cancel?

It is a problem that it is not understood as a cancel action. The Escape key bind is out of context for the menu, so it will close the menu and discard uncommitted changes such as key bind or value change

Metaphysical1:
thanks, I hope you catch syphlis

PR #2985, opened 2021-06-11
Better colour picker :)

Rainerr123:


R3nzTheCodeGOD:
can you share a screenshot ?

danielkrupinski (commented):
I'm not convinced that it's "better" to have a hue wheel instead of a square picker.

nieatty:
> I'm not convinced that it's "better" to have a hue wheel instead of a square picker.

I prefer the square picker, my 2 cents worth.

Smellon69:
> > @Rainerr123
> > image
> > i dont have enough space on the Rainbow tab can u help me?
> 
> stop fucking pasting and learncpp.com

lmfao

PR #3005, opened 2021-06-18
[Bugfix] Fix reading of ColorToggle3 objects from config

mdilai:
Found that "Bomb timer" option doesn't enabled after config loading, although correctly written to config. Found a race condition for `read<value_t::object>`, fixed by moving `static void from_json(const json& j, ColorToggle3& ct)` to `ConfigStruct.h`


danielkrupinski (approved):

danielkrupinski:
Thanks! I think config structs need a recode to prevent such errors in the future.
P.S. That's not a race condition but an implicit cast to base class.

PR #3118, opened 2021-07-19
ViewModel XYZ Changer

blackMasterLord:


KurastikO:
even without osiris my viewmodel commands are messed up.

there's no limit anymore and I can't adjust it properly, dunno why that happens but howwww

wemanzoz:
@blackMasterLord it would be better if we have an option to turn off/on xyz since it overrides your viewmodel in game even without changing a single setting

wemanzoz:
also it stutters your gun for some reason, (when moving left and right)

wemanzoz:
@blackMasterLord please add a checkbox that can make this feature off and on

blackMasterLord:
> @blackMasterLord please add a checkbox that can make this feature off and on

it's done.
i think.

xomarko:
when i enable its so zoomed in lol

xomarko:
image

PR #3121, opened 2021-07-19
Better Watermark

blackMasterLord:


FozzyHvH:
Could I get some help on discord? When enabling the watermark the menu goes quite transparent 
image

PR #3127, opened 2021-07-21
Draw AimBot FOV

blackMasterLord:


floydGG:
This feature affect Aimbot smoothing somehow, i've tried the lowest smoothing value and the highest.
But Aimbot looks snappy af.

danielkrupinski (commented):
It definitely should be implemented inside Aimbot module (Aimbot.cpp)

Frouk3 (changes requested):
Update to the latest Osiris version

Wichtxg:
The hooks.cpp datei is different

Xtendera:
> The hooks.cpp datei is different
> 
> It isn't working

I have a solution for you. It is very easy, just click "Update" and it will work. Just the way you like it.

Link: [Click Me (Very EASY)!!!](https://www.w3schools.com/cpp/)

PR #3137, opened 2021-07-26
Fix visibility check with glow while dead

exploitmafia:
This is just a small issue I noticed with the cheat a few months ago that hasn't been fixed. This seems to fix everything, the problem here occurs when you have glow enabled with separate colors for visibility, and while spectating localplayer can't see the player, but the player you are spectating can. This is kind of hard to explain, but in bot games with fixes the problem completely. If you need further explanation or if I missed something, please leave a comment.

Thanks!

### DIFF
--- a/Osiris/SDK/Entity.cpp
+++ b/Osiris/SDK/Entity.cpp
@@ -114,10 +114,14 @@ bool Entity::visibleTo(Entity* other) noexcept
 {
     assert(isAlive());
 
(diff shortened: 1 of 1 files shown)


You02:
and so you're gay, so uh... umm///...

linuxer9:
This also fixes the problem with visible only ESP if local player is dead, I don't know why this isn't merged yet.

Metaphysical1:
this crashes if you are the T and everyone in your team is dead and bomb planted, when playing competitive, retake or wingman.
there is must be null check missing.

PR #3171, opened 2021-08-06
Between Shots fix

blackMasterLord:


Shuvi-Moro:
I don't think this is a good idea since Between Shots on pistols is a very handy thing

mkot2:
> 
> 
> I don't think this is a good idea since Between Shots on pistols is a very handy thing

This doesn't change anything for pistols, it only fixes it for automatic weapons

PR #3184, opened 2021-08-09
Enable precompiled headers on Linux

mdilai:
Probably this was should be added to #2841 but i forgot about that PR. Anyway, here is little patch to enable compilation of PCH on CMake builds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.13.0)
+cmake_minimum_required(VERSION 3.16.0)
 project(Osiris)
 
 set(CMAKE_C_FLAGS_RELEASE "-O3 -fvisibility=hidden -flto -fno-exceptions -DNDEBUG -Wfatal-errors")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -27,5 +27,6 @@ find_package(OpenGL REQUIRED)
 add_library(Osiris SHARED ${SOURCES})
 
 target_include_directories(Osiris PRIVATE ${SDL2_INCLUDE_DIRS} ${FREETYPE_INCLUDE_DIRS})
+target_precompile_headers(Osiris PRIVATE "${CMAKE_SOURCE_DIR}/Osiris/pch.h")
```
danielkrupinski, line 30: `target_precompile_headers()` requires CMake 3.16 so `cmake_minimum_required()` should be adjusted

danielkrupinski (changes requested):

CMakeLists.txt
mdilai: I didn't think about this. Updated to 3.16.0.

danielkrupinski (approved):

danielkrupinski:
Thanks!

PR #3198, opened 2021-08-16
Improve fake ban

Nep-Timeline:
u don't need reconnect to the server, but only ur team can see it.


b0yk3x:
why? what's the problem with fakeban?

2933968621:
> why? what's the problem with fakeban?

if u like original, u can ignore this change

2933968621:
> why? what's the problem with fakeban?

i think it's better, not fix

b0yk3x:
lol i thought this was issues

PR #2503, opened 2021-01-20
Profile Changer

caiocinel:
Warning: This is a Pull Request without any merge interest, it only serves as a basis for someone who wants to add a ProfileChanger to their build.

The hook method is vfunc, it needs adaptation for the new libraries.

The added files already have full capacity for adding an InventoryChanger / MedalChanger, but I didn't have the patience for that.

If you know how to optimize or extend, feel free to share, that I will try to do the best (or make a pull request on the fork).

In short: it has what it takes to function and nothing else. 

(se você for Brasileiro: Isso é uma pura gambiarra).

Credits: a set of several leaked sources that I found in corsair.wtf;

PR #3229, opened 2021-09-08
add fakePrime

Nep-Timeline:
if not, switch to the match ui
TIPS:this just fake, if u toggle to RANKED MATCH,ur match is still not unranked
but if player in ur room, he see u is a prime player
(linux is not work)

### DIFF
--- a/Osiris/Hacks/Misc.cpp
+++ b/Osiris/Hacks/Misc.cpp
@@ -108,6 +108,7 @@ struct MiscConfig {
     bool killMessage{ false };
     bool nadePredict{ false };
     bool fixTabletSignal{ false };
+    bool fakePrime{ false };
     bool fastPlant{ false };
     bool fastStop{ false };
     bool quickReload{ false };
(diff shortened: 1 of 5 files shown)


SunzBody:
visible for others ?

SunzBody:
good job !

acey1337:
nice one

Callum151800:
It's working...But ranks are missing. Can you update so that ranks can be displayed? (Maybe we can set our rank and how many matches we win)

2933968621:
> It's working...But ranks are missing. Can you update so that ranks can be displayed? (Maybe we can set our rank and how many matches we win)

this just fake, your match is unranked, if you wanna get ranks, you can try use fakeprime and profilechanger #2503

Callum151800 (approved):

ledyyq:
may i ask where is the feature located in the menu?

2933968621:
> may i ask where is the feature located in the menu?
um... misc

PR #3263, opened 2021-09-23
Fix quickReload

R3nzTheCodeGOD:


danielkrupinski (commented):
This does exactly the same.

mdilai:
> This does exactly the same.

Maybe, but now code more consistent. Lets keep Osiris beauty :smile:

PR #3265, opened 2021-09-24
Update Memory::lineGoesThroughSmoke pattern for Linux

mdilai:
OMG, spend almost whole day on this crap but i'll not let linux version of Osiris to die.


danielkrupinski (approved):

danielkrupinski:
Thanks!

PR #3280, opened 2021-09-25
Update memory->insertIntoTree pattern for Linux

mdilai:
Hooks::listLeavesInBox() and Misc::shouldDisableModelOcclusion() was broken for a probably a long time due a silence nature of this fault. Found new signature and corrected it.

P.S. Only now, during pull request creation, i realized that just one byte has been changed so, maybe, its a good candidate to be replaced with "?" symbol for future-proof? I am new to such low-level stuff and don't fully understand when and why it need to be used.

### DIFF
--- a/Osiris/Interfaces.h
+++ b/Osiris/Interfaces.h
@@ -8,6 +8,8 @@
 #include <Windows.h>
 #else
 #include <dlfcn.h>
+
+#include <SDL2/SDL_messagebox.h>
 #endif
 
 #include "SDK/Platform.h"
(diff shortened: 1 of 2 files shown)


mdilai:
Also this could have been detected faster and easier if Linux build can report about patterns/interfaces issues like on Windows. Now it can.

danielkrupinski (commented):
Please split this PR into two:
- message boxes for linux
- Memory::insertIntoTree update

danielkrupinski:
> P.S. Only now, during pull request creation, i realized that just one byte has been changed so, maybe, its a good candidate to be replaced with "?" symbol for future-proof? I am new to such low-level stuff and don't fully understand when and why it need to be used.

Wildcards are used commonly for bytes that might frequently change such as offset in relative long jumps (`E8` opcode, `\xE8????` pattern is very frequent in Memory.cpp). `74` opcode is a near jump instruction so the value you update in this PR is the offset. Near jump offsets change only (mostly) when a function they're in is edited or recompiled. Wildcarding them is a good idea but you have to make sure your pattern is unique to the module (e.g that only one `\x74?\x4C\x8B\x10` exists in `client.dll`).

PR #3316, opened 2021-10-06
Osiris (InventoryChanger extention) - Fake Case Opening

skirtsandflirts:
**This doesn't have to be merged if you don't feel like merging it. I understand if you feel like this code doesn't contribute to the code, or if the feature is unnecessary. If you do decline this though, people can still use it if they feel like it.**

Sorry if my code is bad, I am only just getting back into C++.
I decided that a cool feature would be to add an automated fake case opening.

## What does it do?
Basically, if the feature is enabled, and if you are opening a case through the InventoryChanger, it sends a radio command (Radio.WePlanted, Radio.Flashbang, etc to name a few) and then blanks it out. Under the radio command, we write our own text with colours.

## Positives
- Toggle-able feature.
- This feature works with all cases / capsules. If you open a sticker capsule, it will show in chat. Even with music kit crates!

## Negatives
- The CSGO chat has a restriction on how many characters can fit in one message. Therefor, the shorter your name, the more characters that can fit in the message. If you don't want your skin being cut off the chatbox, set your name shorter. (In this example here when I was testing the code) [https://cdn.discordapp.com/attachments/815602027862163456/894524457169866762/unknown.png](url)
- Purple colour (afaik) doesn't exist, and so we only have Blue, Pink and Red rarities showing in the chat. If you find the Purple colour, you can let me know and I will quickly put that in for you.
- Some things like Stickers might be of an invalid template. Meaning, it might not look legit because some things might have different text. Since I have never seen someone open a sticker capsule (and for it to display in chat), I can only guess what it looks like. If you do have any information on how things are structured in the chat (with what text), let me know and I can gladly change it.

## What have you added?
- Code gets used when a case is opened (from InventoryChanger)
- Created Checkbox in ImGui which toggles the feature on and off.
- Checkbox reads/writes from/to config


wemanzoz:
time to try i guess

skirtsandflirts:
> 
> 
> time to try i guess

I am ass at coding

2933968621:
this is problematic in other languages

skirtsandflirts:
> 
> 
> this is problematic in other languages

Very true. I didn't quite think about this. I don't know how I would add text from a different language (or check it either).

2933968621:
> > this is problematic in other languages
> 
> Very true. I didn't quite think about this. I don't know how I would add text from a different language (or check it either).

i think this is impossible. You can't let other clients see the message in his language

skirtsandflirts:
> 
> 
> can other players see the unboxing if so that's awesome

Only teammates as it uses the Radio command. Look at the downsides for more information.

PR #3324, opened 2021-10-12
Using std::make_unique_for_overwrite() for more efficiency

mdilai:
We don't need to zero-initialize heap memory if we are going to overwrite it immediately. For sure, it have a little sense for stuff like hooks and fonts, changed only for consistency. But it doing a real deal with PlayerAvatar initialization in PlayerData::PlayerData(). With "std::make_unique" i got microstutters on large casual publics everytime when new player connected and new avatar getting cached. With "std::make_unique_for_overwrite()" it's a lot less, almost unnoticeable.

Btw, as we have constant rgbaDataSize i've tried to replace heap allocation with std::array but got extreme stutters on avatars downloading. Probably because lack of move operation for std::array.


skirtsandflirts:
Interesting.

PR #2520, opened 2021-01-23
BlockBot (player blocker)

doggobot-owner:
My first time opening a pull request, if it has any bugs - feel free to leave them here. 
Feel free to use this code for whatever you want.

PR #3326, opened 2021-10-12
BlockBot (player blocker)

E4zily:
A remake of my old request (#2520 ) because i saw a lot of people mentioning it and i see that people want it.
Ported to the latest version of Osiris (by 12.10.2021), use this code as much as you like.


Liam5750:
Daniel is to lazy to press the merge button so never. Sorry, it do be like that.

Metaphysical1:
> Daniel is to lazy to press the merge button so never. Sorry, it do be like that.

that's the beauty of free open source software, anyone can make his own awesome fork and keep it updated with mainstream.
osiris has never been the most feature intensive cheat, but it has basic features that any legit cheat should have and they work well, adding more redundant feature will make the cheat harder to maintain.

E4zily:
this is never getting pushed ;(

PR #3344, opened 2021-10-18
Multiplatform Aimbot::traceToExit() function (AutoWall for Linux)

mdilai:
Reversed traceToExit() function instead of relying on fancy ASM hack on Windows and... nothing at all on Linux. Unfortunately, i was not able to fully abandon usage of memory patterns, because isBreakableEntity() function is hard to reverse as it depend on dynamic casts while Osiris compiled without RTTI support. But if it can be called without workarounds, it's a better to search for it than for traceToExit() memory pattern.
Expected to be multiplatform but tested on Linux only. I am not sure if __FASTCALL is a right calling convention at Osiris/Memory.h line 71.
If will not work on Windows (or just work worse than ASM hack) it can be embraced into #ifdef's to make it linux-only.

### DIFF
--- a/Osiris/Hacks/Aimbot.cpp
+++ b/Osiris/Hacks/Aimbot.cpp
@@ -27,6 +27,11 @@ Vector Aimbot::calculateRelativeAngle(const Vector& source, const Vector& destin
     return ((destination - source).toAngle() - viewAngles).normalize();
 }
 
+/*
+ * "range" and "step" parameters controls distance and accuracy of tracing.
+ * Less "range": more performance
+ * Less "step": more accuracy
+ */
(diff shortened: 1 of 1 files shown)


mdilai:
Added comments about "range" and "step" parameters used by me instead of hardcoding Valve constants. It's useful to be able to play with this parameters but they using Valve default constants by default.

2933968621:
no problem on windows

Metaphysical1:
works great for semirage, only missing hitchance for autoshoot.

mdilai:
> works great for semirage, only missing hitchance for autoshoot.

It will never be accepted but soon i will try to PR legit trigerbot hitchance (which is critically needed for current, almost always missed trigerbot) and you will be able to use your brain and enable it for aimbot too. Anyway, very outdated rewritten semi-rage aimbot (with hitchance, of course) are in my github.

PR #3365, opened 2021-10-26
Linux: Unlock mouse cursor in GUI

mdilai:
Not sure if this nasty thing happened after some of recent updates or i just didn't notice it before, but currently when GUI opened in-game, mouse cursor become locked in middle of screen. I must close GUI, open game menu / console / chat window / etc to get a mouse pointer and only then i could open GUI and actually use it. Got fixed by moving lockCursor hook init outside of #ifdef _WIN32.

Also when cheat has been just injected and automatically opened GUI i often get stuck without any cursor at all until i click twice on Insert. Fixed by moving `ImGui::GetIO().MouseDrawCursor = gui->open;` outside of `if (Misc::isMenuKeyPressed())` scope, so that it draws/hides the cursor directly on GUI opening/closing instead of waiting on keypress.

### DIFF
--- a/Osiris/GUI.cpp
+++ b/Osiris/GUI.cpp
@@ -129,10 +129,10 @@ void GUI::handleToggle() noexcept
         open = !open;
         if (!open)
             interfaces->inputSystem->resetInputState();
-#ifndef _WIN32
-        ImGui::GetIO().MouseDrawCursor = gui->open;
-#endif
     }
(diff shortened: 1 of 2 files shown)


Metaphysical1:
I'm using KDE, the problem it draws 2 cursors if I was on main menu, but this PR fixed it.
thanks

PR #3449, opened 2021-12-13
Add glow thickness

Frouk3:
It's basically glow thickness,i've seen that glow have small outline so i decided to add this, @danielkrupinski i guess you can add this


Frouk3:
Create another pull request because can't update changes

PR #3450, opened 2021-12-14
Update Entity.cpp

MissedShot:
-ShouldSkipAnimationFrame offset fixed


notgoodusename:
Alright, after doing some research, the old offset is ok
The 0x274 offset is used in [DoExtraBoneProcessing](https://github.com/perilouswithadollarsign/cstrike15_src/blob/master/game/client/c_baseanimating.cpp#L3340), by setting it to 0, you skip the call to it
Screenshot
[UC post](https://www.unknowncheats.me/forum/2383502-post8.html) I found which also explains this

PR #3457, opened 2021-12-15
Glow thickness

Frouk3:
Created another one because i wasn't able to update

I've seen that glow outline is small so i decided to change this

@danielkrupinski hey up,will you merge this?


Frouk3:
btw glow thickness applies for all glow effects

PR #3562, opened 2022-02-11
Changes to the autostrafe

ZaUserA:
Some small changes to the osiris' autostrafe to make it directional because I saw people complaining about it (and I also had some free time).


hrna94:
GOOD JOB!

Shuvi-Moro:
> This works 9999/100 works so much better now

What exactly is better about this?

mkot2:
> > This works 9999/100 works so much better now
> 
> What exactly is better about this?

autostrafe now works when you bhop sideways and backwards

jmplpark:
it keeps saying 'identifier vector is undefined'

PR #3586, opened 2022-02-19
Fix Crash

emreires:
SendPacket variable is now in [ebp-0x34] rather than [ebp-0x1C].

### DIFF
--- a/Osiris/Hooks.cpp
+++ b/Osiris/Hooks.cpp
@@ -180,7 +180,7 @@ static bool STDCALL_CONV createMove(LINUX_ARGS(void* thisptr,) float inputSample
         return result;
 
 #ifdef _WIN32
-    bool& sendPacket = *reinterpret_cast<bool*>(*reinterpret_cast<std::uintptr_t*>(FRAME_ADDRESS()) - 0x1C);
+    bool& sendPacket = *reinterpret_cast<bool*>(*reinterpret_cast<std::uintptr_t*>(FRAME_ADDRESS()) - 0x34);
 #else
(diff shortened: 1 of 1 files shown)


Fragrus:
Have you tested this? I corrected this line - but cs:go still crashed when loading onto the map

emreires:
> Have you tested this? I corrected this line - but cs:go still crashed when loading onto the map

Are you using Linux?

Fragrus:
> > Have you tested this? I corrected this line - but cs:go still crashed when loading onto the map
> 
> Are you using Linux?

No Windows. It seems I understood what was the matter, collected it through Osiris.sln, and not through cmake - and it stopped crashing. I will check and write

Fragrus:
Strange, it still crashes through cmake. I don’t know what’s wrong, but the fix works if you collect it through the project file at the root (Osiris.sln). Thanks :)

techzin:
Yeah, fixed.

hrna94:
Is it really working for you guys? After commit and rebuild game dont crash, but osiris is not loaded - cant go to the osiris menu.

2933968621:
i think maybe this will fix the cheat crash
but sendpacket won't work

cryeprecision:
This is definitely not true for Windows (idk about Linux, didn't look at the disassembly).
The value of `bSendPacket` is no longer pushed onto the stack (and restored later), so you'll have to think of something else.

Jonizz:
Still crashing

SefeKHVH821 (approved):

PR #3589, opened 2022-02-22
sendPacket fix for Windows

xor-128:
I tried this only on windows so I enabled only for Windows.


xor-128 (commented):
Clean up

xor-128 (commented):
Clean up - 2

PR #3611, opened 2022-03-15
remove frostbite map check

Metaphysical1:


Patrixsin:
coop_autumn map check is useless too

PR #3811, opened 2022-09-25
Fixed quickReload bug

m3ma110c:
quickreload doesn't work when dropping secondary weapons
https://imgur.com/a/cMRhLWu


rndmlf:
works fine no problems ;)

PR #4012, opened 2023-04-07
Update README.md

Weirdchupacabra:
I changed g++-11 to g++ so the source will actually compile. g++-11 results in an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ Install dependencies:
 
 Configure with CMake:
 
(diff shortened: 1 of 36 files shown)


MissedShot:
G++-11 is for ubuntu/Debian based distros, that ships with lower version of g++ compiler by default, so we need to install and use 11 version manually

Other distros like fedora/arch ships with g++ 12 and above by default, so we need to separate this commands for each distro

PR #4130, opened 2023-08-21
Added Chams backtrack rendering for all previous recorded positions

Kid0h:
I've seen an issue about this already, not too much to say but this is pretty standard in all cheats as rendering only the first position is kind of limiting and hard to utilize.


JannesBonk:
> I've seen an issue about this already, not too much to say but this is pretty standard in all cheats as rendering only the first position is kind of limiting and hard to utilize.

this type of pr was public for quite some time

Xsintashi:
this shit lags for low end pc asf. Make that at least toggable

PR #4174, opened 2023-10-26
Footstep visualization

danielkrupinski:
footstep visualization


fanat1cc:
Maybe do fading circles and also have a visible check so it only shows the circle around corners or dist or smth like that

PR #4194, opened 2023-11-30
Enhance osiris build for windows & linux environment

aa6my:
- make it windows available to run in cmake
- cmake one line build command for windows & linux
- option LINKER:--no-undefined for linux only


danielkrupinski:
- I don't think we should include building with CMake on Windows in the README because it has no advantage over building from VS project and requires additional software (CMake).
- CMake configuration is different from the build command: configuration is done once when changing build type or compiler while the build command is execute multiple times without reconfiguration. These two commands together wastefully perform reconfiguration before every build.
- `if(NOT CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC")` already prevents `LINKER:--no-undefined` under MSVC.

PR #4212, opened 2024-01-07
Fix main menu panel pattern on linux

Jovvik:
7706a32 only updated the windows pattern

xref `"file://{resources}/layout/mainmenu.xml"` in case this happens again


danielkrupinski (commented):
New pattern should work both for current and previous game version in order to make it less likely to break in the future.

PR #4300, opened 2024-06-03
added a way to hide tpid in readme

tranarchy:


3152460b:
微信图片_20240606233506
What to do

5chanhackeranonymis:
This is stupid; you're better off just using a better injection method. The best way for now is probably replacing a .so cs2 is linked to.



> 微信图片_20240606233506 What to do

Silly skid can't read

5chanhackeranonymis:
cat /proc/$(pidof cs2)/maps | grep -i libOsiris.so

LoongArum:
This just use less, if valve really care about cheater on Linux,there are lots of way.. like use ptrace to debug it self, so it can detect gdb ass well. value just don't care..., we just use this little toy until by VAC dude.

LoongArum:
> 微信图片_20240606233506 What to do

使用 Linux,另外不要用微信截图，我能找到你的物理位置。

Map1eUM:
> 微信图片_20240606233506 What to do

If you are on windows you don't need ATT

LoongArum:
> > 微信图片_20240606233506 What to do
> 
> If you are on windows you don't need ATT

So strange, Who posted an error pic in the random pull request, but Lucy just saw and pointed out the problem. You are so kind!

PR #4625, opened 2026-04-24
Fix for the latest game update

lZiMUl:


ThePrimeDev:
Please, do not open a pull request with the patterns in my fork, because 1.) It's not fully updated, some patterns are still missing, 2.) Some patterns of mine might be less optimal then the pattern generator that Daniel usually uses, this way if theese patterns get into the main repository of the cheat, they need to be updated more often.

In any other way, feel free to use the stuff in my fork, just let's not bother Daniel unnecessarily with theese kind of stuff.

PR #4658, opened 2026-06-13
fix linux pattern after latest game update

KebabLord:
all functionalities seem to work with this pattern update
though i dunno what the new patterns on windows are, this PR only updates linux.


Gorkido:
the menu doesn't show up.

KebabLord:
@Gorkido works for me even after receiving 2 updates, how do you inject the game? I recommend this:
```bash
sudo gdb -q -p "$(pidof cs2)" \
  -ex 'call (void*)dlopen("/path/to/libOsiris.so", 2)' \
  -ex 'continue' \
  -ex 'detach' \
  -ex 'quit'
```

Gorkido:
@KebabLord the injection seems to be working, but i don't get the menu.

```
--Type <RET> for more, q to quit, c to continue without paging--
-
-
- (child processes)
-
-
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
0x00007f6284ab58b2 in ?? ()
   from target:/usr/lib/pressure-vessel/overrides/lib/x86_64-linux-gnu/libc.so.6
$1 = (void *) 0x0
Continuing.
```

I'm on cachyos kernel 7.1.1-2-cachyos
Also, if you or anyone finds the solution relating to this problem in [issues](https://github.com/danielkrupinski/Osiris/issues), please mention. thanks.

dark-ioux:
why u close pull request

PR #4681, opened 2026-07-21
Update: Planted c4 patterns for windows

iiNothh:
Trust me, Daniel—these are some of the most reliable patterns you'll find.


danielkrupinski (commented):
In the first pattern replacing one byte with a wildcard is enough, and the second pattern can be shorter. Nevertheless, thanks for the effort.

